PR #23, opened 2021-11-08
Replace OIDC introspection with JWKS

paulius-valiunas:


packages/service/package.json
```diff
@@ -23,18 +23,24 @@
     "url": "https://github.com/iTwin/auth-clients"
   },
   "dependencies": {
+    "@itwin/core-common": "nightly",
```
calebmshafer, line 26: Why do we need this added?

paulius-valiunas: somehow this got in here with a merge. Let me fix it...

calebmshafer (approved):

PR #28, opened 2021-11-17
work in progress updates to the Electron readme

calebmshafer:


packages/electron/README.md
```diff
@@ -4,18 +4,24 @@ Copyright © Bentley Systems, Incorporated. All rights reserved. See LICENSE.md
...
 
-3. The backend client needs to setup listeners via the `ipcMain` object from Electron. The listeners should use the exact same channel names that the frontend uses. The listeners will response to requests to sign-in, sign-out, and retrieve the access token. When a token changes on the backend then it becomes necessary to send messages to the frontend unprovoked which cannot be done with the `ipcMain` object. Instead it must be done by accessing the window object's 'webContents' which has a `send` method that allows you to send data over a specific IPC channel name. An example of all this can be seen in [BackendClient.ts](./src/backend/BackendClient.ts).
+The list above is in the order that each of them need to be handled. The preload script is first as it is used on startup of an instance of Electron, the backend client is order of the above is important due to the order
```
pbell97, line 27: Just pointing out that this paragraph/thought isn't finished.

calebmshafer: just removed this sentence for now

README.md
```diff
@@ -1,4 +1,4 @@
-# iTwin.js
+# iTwin.js Authentication clients
```
pbell97, line 1: Shouldn't this be "Authorization"?

packages/electron/README.md
```diff
@@ -4,18 +4,22 @@ Copyright © Bentley Systems, Incorporated. All rights reserved. See LICENSE.md
...
 
 ## How to setup
 
-The Electron frontend and backend auth clients communicate via Electron's IPC Electron in order to exchange login and access token information. The Electron must be setup correctly for it to work. There are 3 general steps to set this up:
+An Electron application must follow a few setup steps in order to consume and use these classes correctly.
+
+1. The Electron preload script (required for [context isolation](https://www.electronjs.org/docs/latest/tutorial/context-isolation)) must create an implementation of `ITwinElectronApi` to allow the use of the `ipcRenderer` object provided by Electron. The implementation must be exposed in the window object so that the renderer process has a means of accessing the Electron IPC. This is required because the renderer process does not naturally have access to the `ipcRenderer` object to interact with Electron IPC. An example of this can be seen in [ElectronPreload.ts](./src/renderer/ElectronPreload.ts).
```
pbell97, line 19: This may be me being a little nitpicky but we could probably combine two of the later sentences that say almost the same thing.

```suggestion
1. The Electron preload script (required for [context isolation](https://www.electronjs.org/docs/latest/tutorial/context-isolation)) must create an implementation of `ITwinElectronApi` to allow the use of the `ipcRenderer` object provided by Electron. The implementation must be exposed in the window object so that the renderer process has a means of accessing the Electron IPC since it doesn't naturally have access to the `ipcRenderer`. An example of this can be seen in [ElectronPreload.ts](./src/renderer/ElectronPreload.ts).
```

packages/electron/README.md
```diff
@@ -4,18 +4,22 @@ Copyright © Bentley Systems, Incorporated. All rights reserved. See LICENSE.md
...
 
-2. The frontend client needs to create and use a basic class that interacts with the `ITwinElectronApi` that is referenced from the window object. It should setup not only listeners to react when a token is changed on the backend but also methods that will invoke the backend when a sign-in or sign-out needs to happen. An example of this class (`ElectronAuthIPC`) and how the frontend uses it can be seen in [FrontendClient.ts](./src/frontend/FrontendClient.ts).
+2. `ElectronMainAuthorization` needs to setup listeners via the `ipcMain` object from Electron. The listeners should use the exact same channel names that the renderer process uses. The listeners will response to requests to sign-in, sign-out, and retrieve the access token. When a token changes in the main process then it becomes necessary to send messages to the renderer process client unprovoked which cannot be done with the `ipcMain` object. Instead it must be done by accessing the window object's 'webContents' which has a `send` method that allows you to send data over a specific IPC channel name. An example of all this can be seen in the main [Client.ts](./src/main/Client.ts).
```
pbell97, line 23:
```suggestion
2. `ElectronMainAuthorization` needs to setup listeners via the `ipcMain` object from Electron. The listeners should use the exact same channel names that the renderer process uses. The listeners will respond to requests to sign-in, sign-out, and retrieve the access token. When a token changes in the main process then it becomes necessary to send messages to the renderer process client unprovoked which cannot be done with the `ipcMain` object. Instead it must be done by accessing the window object's 'webContents' which has a `send` method that allows you to send data over a specific IPC channel name. An example of all this can be seen in the main [Client.ts](./src/main/Client.ts).
```

pbell97 (approved):
Minor suggestions but overall looks good to go.

README.md
calebmshafer: good call, fixed

packages/electron/README.md
calebmshafer: That sounds better to me. I took the suggestion.

PR #44, opened 2021-12-06
Add linux support

pbell97:
Removed linux restrictions are updated the readme with information on how to run headless.

TODO: Decide how we want to handle setting up pipelines to be able to handle the correct headless environment.

### DIFF
--- a/packages/electron/README.md
+++ b/packages/electron/README.md
@@ -18,8 +18,21 @@ An Electron application must follow a few setup steps in order to consume and us
 
 1. The Electron preload script (required for [context isolation](https://www.electronjs.org/docs/latest/tutorial/context-isolation)) must create an implementation of `ITwinElectronApi` to allow the use of the `ipcRenderer` object provided by Electron. The implementation must be exposed in the window object so that the renderer process has a means of accessing the Electron IPC since it doesn't naturally have access to the `ipcRenderer`. An example of this can be seen in [ElectronPreload.ts](./src/renderer/ElectronPreload.ts).
 
-    > Note: The `@itwin/core-electron` package handles this for any iTwin.js application working with an iModel so this step can be skipped if you're starting you app using `ElectronHost`.
+    > Note: The `@itwin/core-electron` package handles this for any iTwin.js application working with an iModel so this step can be skipped if you're starting your app using `ElectronHost`.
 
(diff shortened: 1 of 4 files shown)


mattbjordan:
The only issues with this PR when I started were that the Linux tests were passing. They need some extra setup to work correctly with Keytar. I followed [Patrick's documentation](https://github.com/iTwin/auth-clients/blob/de4bf9a87b4af4a05c1c1ffb66f9ca8b7e843549/packages/electron/README.md#linux-usage) (added in this PR) to get the tests passing.

mattbjordan:
The Auth-Clients-CI pipeline is a Github Action workflow as opposed to being an Azure Pipeline. I see that there exists an unused Azure Pipeline .yml that does performs the same tasks plus also publishes tests results and packages. 

This PR was started over a year ago, so I'm questioning if using the Github Action is is outdated or not? If we want to switch to the Azure Pipeline, I'll do that in a separate PR. Though, I did try to add the relevant fixes from this PR to the Azure yaml file already.
 
@calebmshafer @aruniverse

packages/electron/README.md
```diff
@@ -18,8 +18,20 @@ An Electron application must follow a few setup steps in order to consume and us
...
+
+## Linux Usage
+
+`ElectronMainAuthorization` uses the node package [Keytar](https://www.npmjs.com/package/keytar) to securely save refresh tokens to disk. This allows the client to automatically sign-in and receive a new access token between sessions. Keytar does this by using different native secure storage solutions depending on the operating system, and it uses libsecret on linux. In order to use keytar on linux, specifically Debian/Ubuntu, `libsecret-1-dev` must be installed.
+
+If keytar is being used in a headless environment additional steps need to be taken. The following packages will need to be installed:
+
+- libsecret-1-dev
```
aruniverse, line 33: id wrap each one of these in backticks since you did so above

ben-polinsky (approved):
Not sure about the ADO vs Github Actions question, but otherwise looks good.

mattbjordan:
> i'm ok with just using gh pipelines for testing, so you can delete the unused ado pipeline. good work on this!

I'm going to keep the file.

I found that [the Azure pipeline was set some time ago to only be triggered manually](https://github.com/iTwin/auth-clients/pull/42). Plus, I didn't realize that the [iTwin.Auth-Clients.Publish pipeline on ADO](https://dev.azure.com/bentleycs/iModelTechnologies/_build?definitionId=6189) has the manual runs of the [ci.yml](https://github.com/iTwin/auth-clients/blob/15c89eb779db16882a3afaad26a97bb7af731da9/common/config/azure-pipelines/ci.yml) file. From just looking at the file, I wouldn't have expected that to be the pipeline name, so I didn't realize it was still used.

PR #45, opened 2021-12-15
pin deps to core rc

aruniverse:


pbell97 (approved):

calebmshafer (approved):
we'll need to bump the version of these packages too.

PR #47, opened 2021-12-15
Add silent electron signin

pbell97:
Due to problems with accessing electron packages before Electron is fully setup, the initialize and create methods were removed.

A new signInSilent method was added in its place to initialize when needed.


common/changes/@itwin/electron-authorization/patrick-AddSilentSignin_2021-12-15-22-19.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/electron-authorization",
+      "comment": "Replaced initialize with silent signin",
+      "type": "minor"
```
pbell97, line 6: I guess technically this would be a 'major' change since its breaking the API, no? But also not sure if we want to bump it up to 1.0 yet before official 3.0 release.

calebmshafer: Yea, a minor bump is fine since we haven't moved to 1.0 yet.

packages/electron/src/main/Client.ts
```diff
@@ -271,6 +251,24 @@ export class ElectronMainAuthorization implements AuthorizationClient {
     await authorizationHandler.performAuthorizationRequest(this._configuration, authorizationRequest);
   }
 
+  /**
+   * Attempts a silent sign in with the authorization provider
+   * @throws [Error] If the silent sign in fails
```
calebmshafer, line 256: Can you remove this since it doesn't throw?

pbell97: Removed

packages/electron/src/main/Client.ts
```diff
@@ -76,7 +76,7 @@ export class ElectronMainAuthorization implements AuthorizationClient {
   public get tokenStore() { return this._tokenStore; }
   private static _defaultRequestOptionsProvider: DefaultRequestOptionsProvider;
 
-  private constructor(config: ElectronMainAuthorizationConfiguration) {
+  public constructor(config: ElectronMainAuthorizationConfiguration) {
```
aruniverse, line 79: The constructor is kinda weird to me, and could be optimized imo.

1. This will never throw like you expect, see this [playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIBsILFA9iAWTlAEEBXMAC1ymAC84xh8BhfGYAczKkeZGQBvALAAoZMmABnKWWgBVKOgD8ALmRScoTgG4xEqBDgATfOgCeyQ8eCHsi4Go1aQu-VaOmQF5AnTAIcABJY3VNWlc9cQ8TM0spBFwABwgwlzdow1jvSwgADyTbcwAhMhgYaCcQMgBbACNoKIBfMTE-OBk0TGw8QmIQcioaej58IXcAegnkJLxIbAhjZAB9RCQZABVcAGtA9RIEdakt3YEAXmQAIkuoiSSyOv8EX3YudQwsHHwiUgpqWgYTFYr24vCBIFuMweT2Q+UKhgAynAKmBLBcAGwABmQACpkABGTE6ZBTDwwQxSSjIMA7QIE7E1UAUCBSZANGA0FBZdCwgq2UYCJg1CDue6PYDPHg8i6XShgMBJKSqKbAGpSAB0DXAmHM6sSNRuouhEucfGexgg8DI6DAACVFrZPg5kDK5QrlRN0LgEHB0NRNKoAMyYkMTKRcECgAC0PvQ6DqiG2huipNmwAAbowUMtEiAOKCBfs-sNAfwEdB0xKIGw81weALkAAfZBkEAWjggRaQ1O0TOQFY0072qRJfBSVLIE6BYejkDjpsttuW0BdybTNN97ODwIImmGJwfHr4KcgXecyEbrMrOGOqQkMBOAAiWe76+Nz04EDA1Npp73EAACgASiEDwwB4QVKGkdVlm3P9z2QFoU3XXsr00M0VnbOBrTtCAAEc5E0AB5JJwSkAAFPBKwtKB1EfS1sJte0CJZMASLIyjcGoxpWmiMUYVzcIyGwGgANzfN3m6L4+l+IYAQFGt83rcEQJEaIJCoaDxK4F0XlrDJ3AkYAYGQACAEJNI1bTOHVBJkggdVQD8MgLSkADLlwcp-E7VZDhZKRLiAoC1IkULqSgqyQV0kKwtC9V4ssvUQQAGkM2K7JSdQAAMABJBES6zbMSFImmQTyYG8iBfKOLLUvU0KmkhCQkLS4zTIsiKkv0oC0o0yg8AAd2QTshtQKA8CgdyCDITQNBSBBjMsOBkEzfwlms5T+GQAbKDpUBgCYX16G0LpPl6H4BmLeTwUCqI0swb9ZmXPJdOuJq9Nm7C5IOtFhogIb5FtAAZACCpBRyZDkKBFB5ZRlHC6CpSA962oAp6OBegAyTHkA6rTwekWQFCUVTepmQwMd09HgBes46auC10yjS5kHhy48LgZnkHUam8neiQvv+H71X9MAQDgYUqYpmnkAAamQQXhlREXcE0cXhXepCwsSqVdMV2hlf6y11UMJJ0EQQCJgAHQmHKJmSq5bt4sLUbB-T1RvKASjKCooB6+q+ugz2ICRFFfrd-MPb5L3SnKHjRDS0lEtg38z0MXSRtOo8QBPNPAIjrg9X8QIwBCZH3Bavj31hBBqGA8Z6sE3BMHVL1OHctNwECiuxErwTv0FjP-qz6SLsGIXS3wACYr8AJglCR26okDKJ2uXv-cFj3a9wYCgA)
2. You should probably be checking that clientId and scope are defined, not the config object.
3. Throw earlier on in the constructor.
4. Do you really need to save config as member var of the class? Youre already extracting a bunch of the properties of the config obj, and saving them as member vars. Some of those properties of the config object passed in will be stale, eg youre setting url, but never updating config.issurUrl

pbell97: That's a good point, I can move the check to the top of the constructor and check specifically for clientId and scope since they are the only two required properties. 

I agree it is a little weird to store both the config and several of its properties as member vars, we would also have to add three more member variables (scope, clientId, and redirectUrl) just to get rid of the config. I guess it would be worth it?

@calebmshafer You mentioned you were doing some refactoring of this class in another branch, are you changes related to these possible ones Arun is suggesting?

aruniverse: yea looking at it, i dont think you need to save config at all, could just save the clientid & scope

calebmshafer: Yes, some of my changes do address those concerns. Let me push it up now and give some more detail.

aruniverse: ah yes, youre right the redirectUrl as well

calebmshafer: Here's the PR I started and need to finish up #48. It has some more substantial public API implications as well.

calebmshafer: I think that throw below is entirely unnecessary though. The param to the method ensure it will always contain at an object... Lets go ahead and remove line 90 and 91 in this PR. Then the other clean up can happen in the one I have open.

aruniverse: Youg guys should probably throw if config.clientid and config.scopes are empty strings

calebmshafer: I was debating that and throwing in constructors never really works so I was going to try and avoid it if possible. It is required on the interface so at least it won't be undefined but you're right they could potentially send in empty strings.

pbell97: Yeah I think checking to make sure they aren't empty strings is a good idea, I made an update for that and forgot to push it until just now.

calebmshafer (approved):

aruniverse (approved):
Changes look fine, i guess more refactoring will occur in calebs pr

PR #48, opened 2021-12-16
Update the Electron Auth API

calebmshafer:
The current Electron authorization API exposes quite a few things that it doesn't need to or shouldn't.

In addition to that, the passed in configuration could potentially be manipulated after the instance is created so we want to avoid that.


aruniverse:
closing this in favor of #52

PR #52, opened 2022-01-24
Release for 3.0

aruniverse:
Bump deps to 3.0


pbell97 (approved):

aruniverse:
@calebmshafer when you get a chance, please confirm its ok to leave itwin/core-bentley as a direct dep.

pbell97 (approved):
Looks good, thanks for cleaning up alot of that Electron client too.

evelynpreslar-bentley (approved):

packages/electron/src/main/Client.ts
```diff
@@ -127,33 +135,38 @@ export class ElectronMainAuthorization implements AuthorizationClient {
...
-   * Augments request options with defaults returned by the DefaultRequestOptionsProvider.
-   * @note The options passed in by clients override any defaults where necessary.
-   * @param options Options the caller wants to augment with the defaults.
-   * @returns Promise resolves after the defaults are setup.
-   */
-  protected async setupOptionDefaults(options: RequestOptions): Promise<void> {
-    if (!ElectronMainAuthorization._defaultRequestOptionsProvider)
-      ElectronMainAuthorization._defaultRequestOptionsProvider = new DefaultRequestOptionsProvider();
-    return ElectronMainAuthorization._defaultRequestOptionsProvider.assignOptions(options);
+  public static readonly onUserStateChanged = new BeEvent<
```
calebmshafer, line 152: gross formatting, I assume it's prettier :)

rush.json
```diff
@@ -5,14 +5,12 @@
   "nodeSupportedVersionRange": ">=12.17.0 <17.0.0",
   "projectFolderMinDepth": 2,
   "projectFolderMaxDepth": 3,
-  "ensureConsistentVersions": true,
+  "ensureConsistentVersions": false,
```
calebmshafer, line 8: what did this conflict with?

aruniverse: i had to bump the electron pkgs separately because it was still referencing itwin-client.
wouldve reverted this change, but i thought it made sense here, we should be able to support different core vers for each of the pkgs, and bump them as needed

packages/electron/src/main/Client.ts
aruniverse: sigh, yea... reverted.

rush.json
calebmshafer: Oka, that should be okay then.

calebmshafer (approved):

PR #54, opened 2022-01-24
Update core dependencies

aruniverse:
core-bentley should be a dev+peer dependency


pbell97 (approved):
Looks like you need a `rush update`

calebmshafer (approved):
It's really unfortunate that we need to do this just for logging and events...

PR #60, opened 2022-02-01
cast to resolve type error between union type

aruniverse:


packages/service/src/introspection/IntrospectionClient.ts
```diff
@@ -79,10 +79,10 @@ export class IntrospectionClient {
   }
 
   private async validateToken(accessToken: string): Promise<IntrospectionResponse> {
-    const decoded = jwt.decode(accessToken, { complete: true });
+    const decoded = jwt.decode(accessToken, { complete: true, json: true });
     if (!decoded)
       throw new Error("Failed to decode JWT");
-    const { payload, header } = decoded;
+    const { payload, header } = decoded as { payload: jwt.JwtPayload, header: jwt.JwtHeader };
```
calebmshafer, line 85: what was the exact issue before? Was `decode` not returning as an object with the payload and header props?

aruniverse: for some reason ts couldnt tell whether payload was a string or a JwtPayload

aruniverse: image

calebmshafer: okay, just tested.

calebmshafer (approved):

PR #61, opened 2022-02-02
bump keytar version to address vulnerabilities of transitive dependencies

evelynpreslar-bentley:


aruniverse (changes requested):
rush change minor

aruniverse (approved):

calebmshafer (approved):

PR #63, opened 2022-02-08
Update how the token prefix is removed

calebmshafer:
The current handling resulted in a difficult to parse error when attempting to get the `substr` of a string with length of 0. Updated the handling to use split instead of `substr`.


aruniverse (approved):

calebmshafer:
@evelynpreslar-bentley mind reviewing the PR?

evelynpreslar-bentley (approved):

PR #64, opened 2022-02-19
Added more OAuth providers

arvindsvenkat:
Added changes to support AzureAD and Authing using the oidc-signin-tool


calebmshafer:
> Does anyone know how to resolve the license/cla check? I changed the git config user email and the last commit seems to be verified by CLA. I'm not sure what to do about the previous commits that don't have the correct username associated.

@arvindsvenkat looks like you committed using the your work email. I'd rewrite those commits to make sure they're using the GitHub noreply email.

PR #66, opened 2022-03-09
Add integration test pipeline

calebmshafer:


mattbjordan:
I have the integration tests succeeding and running via Azure pipelines.

I commented out the OIDC Authing test as the authority url is not longer working, and from my understanding is essentially outdated. I do not know the current or future plans for our authentication services so I decided to only comment out the test instead of removing it outright in case it is ever to be revisited.

@calebmshafer @ben-polinsky

packages/oidc-signin-tool/src/test-integration/basic.test.ts
```diff
@@ -165,17 +165,18 @@ describe("Authing and AzureAD (#integration)", () => {
...
-      throw new Error("Could not find IMJS_TEST_AUTHING_USER_PASSWORD");
-
-    const validUser = {
-      email: process.env.IMJS_TEST_AUTHING_USER_NAME,
-      password: process.env.IMJS_TEST_AUTHING_USER_PASSWORD,
-    };
-    const token = await getTestAccessToken(authingOidcConfig, validUser);
-    assert.exists(token);
-  });
+  /** The authing authority does not work anymore. **/
```
ben-polinsky, line 168: This code should be removed. 

If we need to test what this was testing, then it should be rewritten.

common/config/azure-pipelines/integration-test.yaml
```diff
@@ -0,0 +1,89 @@
...
+      Write-Host "##vso[task.setvariable variable=IMJS_TEST_SUPER_MANAGER_USER_PASSWORD;]$Env:test_super_manager_password"
+      Write-Host "##vso[task.setvariable variable=IMJS_TEST_SUPER_USER_PASSWORD;]$Env:test_super_user_password"
+      Write-Host "##vso[task.setvariable variable=IMJS_TEST_MANAGER_USER_PASSWORD;]$Env:test_manager_user_password"
+      Write-Host "##vso[task.setvariable variable=IMJS_TEST_REGULAR_USER_PASSWORD;]$Env:test_reg_password"
+      Write-Host "##vso[task.setvariable variable=IMJS_TEST_AZUREAD_USER_PASSWORD;]$Env:test_azuread_user_password"
+    displayName: 'Add config to environment'
+    env:
+      test_super_manager_password: $(IMJS_SUPER_MANAGER_USER_PASSWORD)
+      test_super_user_password: $(IMJS_SUPER_USER_PASSWORD)
+      test_manager_user_password: $(IMJS_MANAGER_USER_PASSWORD)
```
ben-polinsky, line 62: Looks to me like this is not used in the test. If not, this and any other unused variables should be removed to avoid confusion.

ben-polinsky (approved):
I would still like to understand what exactly the test we got rid of does and if it is valuable

common/config/azure-pipelines/integration-test.yaml
mattbjordan: Those values are read by the [TestUsers](https://github.com/iTwin/auth-clients/blob/main/packages/oidc-signin-tool/src/TestUsers.ts) class and used by a test on [line 99 of basic.test.ts](https://github.com/iTwin/auth-clients/blob/832ba430bb52ede8c31495a56b4be39ab5d3e36b/packages/oidc-signin-tool/src/test-integration/basic.test.ts#L99).

ben-polinsky: Missed that - thanks

packages/oidc-signin-tool/src/test-integration/basic.test.ts
mattbjordan: Okay, I've removed it then. The service that this was testing doesn't work anymore as I mentioned in my comment above.

ben-polinsky (approved):

PR #69, opened 2022-04-04
Add simple command-line authentication client for devtools

mgooding:


packages/node-cli/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@itwin/node-cli-authorization",
+  "version": "0.8.3",
+  "main": "lib/cjs/Client.js",
+  "types": "lib/cjs/Client.d.ts",
+  "description": "Node.s command-line authorization client for iTwin platform",
```
calebmshafer, line 6: `Node.s` -> `Node.js`

calebmshafer (commented):
Still reviewing the full setup but an initial pass it looks good. I'll do a deeper dive tomorrow.

packages/node-cli/package.json
mgooding: fixed

mgooding:
I'd like to have tests for the actual sign-in flow, but I'm going to wait to leverage some work Raman is doing in this area.

packages/node-cli/src/Client.ts
```diff
@@ -0,0 +1,382 @@
...
+/**
+ * Client configuration to generate OIDC/OAuth tokens for command-line applications
+ * @public
+ */
+export interface NodeCliAuthorizationConfiguration {
+  /** The OAuth token issuer URL. Defaults to Bentley's auth URL if undefined. */
+  readonly issuerUrl?: string;
+  /**
+   * Upon signing in, the client application receives a response from the Bentley IMS OIDC/OAuth2 provider at this URI
+   * For this client, must start with `http://localhost:${redirectPort}` or `https://localhost:${redirectPort}`
```
calebmshafer, line 41: Does an `https` redirect actually work with the current setup?

packages/node-cli/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Copyright © Bentley Systems, Incorporated. All rights reserved. See LICENSE.md for license terms and full copyright notice.
+
+## Description
+
+The __@itwin/node-cli-authorization__ package contains a Node.js command-line based client for authorization with the iTwin Platform by default and is configurable to work with any OAuth2.0 based provider.
+
+## How it works
+
+The OAuth2.0 workflow used in this package is Authorization Code + PKCE, for more information about the flow please visit the [Authorization Overview Page](https://developer.bentley.com/apis/overview/authorization/#authorizesinglepageapplicationsspaanddesktopmobileapplicationsnative).
```
calebmshafer, line 11: Can you mention that they'll want to create a "native" client in order ot use this?

packages/node-cli/src/Client.ts
```diff
@@ -0,0 +1,382 @@
...
+}
+
+/**
+ * @internal
+ */
+export class BakedAuthorizationConfiguration {
+  public readonly clientId: string;
+  public readonly scopes: string;
+  public readonly issuerUrl: string;
+  public readonly redirectUri: string = "http://localhost:3000/signin-callback";
```
calebmshafer, line 300: We should mention the default since it is optional.

packages/node-cli/src/Client.ts
```diff
@@ -0,0 +1,382 @@
...
+  public readonly expiryBuffer: number = 60 * 10; // refresh token 10 minutes before real expiration time;
+
+  public constructor(config: NodeCliAuthorizationConfiguration) {
+    if (!config.clientId || !config.scope)
+      throw new BentleyError(AuthStatus.Error, "Must specify a valid configuration with a clientId and scope when initializing NodeCliAuthorizationClient");
```
calebmshafer, line 305: This throw will get lost at this stage since it's done in the constructor of the client and this class.

packages/node-cli/src/Client.ts
```diff
@@ -0,0 +1,382 @@
...
+  }
+
+  /**
+   * Returns a promise that resolves to the AccessToken of the currently authorized user.
+   * The AccessToken is refreshed as necessary.
+   * If signIn hasn't been called, the AccessToken will remain empty.
+   */
+  public async getAccessToken(): Promise<AccessToken> {
+    if (!this._configuration) {
+      Logger.logTrace(NODE_CLI_AUTH_LOGGER_CATEGORY, "getAccessToken called but not initialized. Call signIn first");
```
calebmshafer, line 89: I'd log this as an error.

packages/node-cli/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "rebuild": "npm run clean && npm run build"
+  },
+  "author": {
+    "name": "Bentley Systems, Inc.",
+    "url": "http://www.bentley.com"
+  },
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/iTwin/auth-clients"
```
calebmshafer, line 25: can you put direct link to this folder? The link in npm will then guide you directly to the sub folder in that case.

packages/node-cli/src/Client.ts
mgooding: No, it wouldn't - I'll remove that. (We should do the same in DesktopAuthorizationClient when that gets cleaned up, I copied from there).

packages/node-cli/src/Client.ts
mgooding: Fair enough - I wasn't sure if it was unexpected for IModelHost to possibly call this before the client is set up.

packages/node-cli/src/Client.ts
mgooding: I don't understand this comment - this works as expected in my testing with imodel-unity-example. If I don't specify clientId or scope there, the exception seems to bubble up properly.

mgooding:
@calebmshafer - everything you mentioned should now be addressed, except for the baked config class throwing in constructor.

packages/node-cli/src/Client.ts
calebmshafer: ah okay, maybe I'm mis-remembering javascript semantics here but I thought that when you threw in a constructor it didn't break out correctly so that you could use a try/catch. I'd need to do some playing around to be sure. However, I see this is how it's setup in the electron auth so seems fine to keep it as is.

calebmshafer (approved):
LGTM!

PR #73, opened 2022-04-27
Include issuerUrl in TokenStore key

mgooding:
Incorporate issuerUrl into TokenStore key to avoid conflicts between credentials that share a clientId but were generated from different issuing authorities. This is a common problem for Bentley when testing between QA and PROD.

Also shortened the "iTwinJs" boilerplate to make the entries more readable in Credential Manager, etc.

### DIFF
--- a/packages/node-cli/package.json
+++ b/packages/node-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/node-cli-authorization",
-  "version": "0.9.1",
+  "version": "0.9.0",
   "main": "lib/cjs/Client.js",
   "types": "lib/cjs/Client.d.ts",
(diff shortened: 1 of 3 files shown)


packages/node-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/node-cli-authorization",
-  "version": "0.9.0",
+  "version": "0.9.1",
```
calebmshafer, line 3: No need to manually update the version.

ramanujam-raman (approved):

evelynpreslar-bentley (approved):

calebmshafer (approved):

PR #74, opened 2022-04-29
Better electron token store key

evelynpreslar-bentley:
Followup to #73 to replicate same functionality in Electron client

### DIFF
--- a/packages/electron/src/main/Client.ts
+++ b/packages/electron/src/main/Client.ts
@@ -11,12 +11,14 @@
 // cSpell:ignore openid appauth signin Pkce Signout
 /* eslint-disable @typescript-eslint/naming-convention */
 
-import type { AccessToken} from "@itwin/core-bentley";
+import type { AccessToken } from "@itwin/core-bentley";
 import { assert, AuthStatus, BeEvent, BentleyError, Logger } from "@itwin/core-bentley";
 import type { AuthorizationClient } from "@itwin/core-common";
 import type {
-  AuthorizationError, AuthorizationRequestJson, AuthorizationResponse, RevokeTokenRequestJson, StringMap, TokenRequestHandler, TokenRequestJson, TokenResponse} from "@openid/appauth";
-import { AuthorizationNotifier, AuthorizationRequest, AuthorizationServiceConfiguration,
+  AuthorizationError, AuthorizationRequestJson, AuthorizationResponse, RevokeTokenRequestJson, StringMap, TokenRequestHandler, TokenRequestJson, TokenResponse
+} from "@openid/appauth";
+import {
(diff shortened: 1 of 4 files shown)


calebmshafer (approved):

packages/electron/src/main/Client.ts
```diff
@@ -105,7 +107,8 @@ export class ElectronMainAuthorization implements AuthorizationClient {
     if (config.redirectUri) this._redirectUri = config.redirectUri;
     if (config.expiryBuffer) this._expiryBuffer = config.expiryBuffer;
 
-    this._tokenStore = new ElectronTokenStore(config.clientId);
+    const appStorageKey = `iTwinJs:${this._clientId}:${this._issuerUrl}`;
```
mgooding, line 110: This isn't quite the same as the CLI client... one sec and I'll post more. Leaving a comment to keep the PR from resolving.

mgooding: See https://github.com/iTwin/auth-clients/blob/main/packages/electron/src/main/TokenStore.ts#L23

I think the PR should also change the ElectronTokenStore constructor to use the `appStorageKey` as-is.

evelynpreslar-bentley: Ah, that's where the hardcoded string is. I was wondering why the tests were failing.

It would be nice if the two clients just shared a TokenStore class. The two are identical except for the appStorageKey (which I'll change) and the presence of the `delete` function.

Is there a reason why they're separate aside from not wanting another package?

mgooding: > Is there a reason why they're separate aside from not wanting another package?

No, I think ideally much more of this code would be shared. I'd initially hoped to unify most of the node-cli and electron code (as well as Raman's sign-in logic for the vscode plugin) but we found the vscode plugin requirements diverged enough to give up on that. After that, I found it was sufficiently tricky for me to understand and untangle the electron code enough to make node-cli something I could maintain.

I still think it would be good to refactor the electron package to make it dependent on node-cli, but it's not a project I plan to take on.

evelynpreslar-bentley: I could be the one to take that on, but for this PR, I'll just kick the can and keep things separate.

Because the Electron tests modify the token store directly, I had to [duplicate the token store initialization code](https://github.com/iTwin/auth-clients/pull/74/commits/e63e8a47ede5e150fe8ef54fc8bd5ea75dcdff90) into the tests since the store itself doesn't do it anymore.

I was thinking of using a static function to avoid duplicating code, but I'm not sure if that's any better just to accommodate tests.

mgooding: Sounds good. Thanks!

PR #75, opened 2022-06-15
Incorporate scopes into caching key for credentials

mgooding:
Ensure we only use cached credentials that match client ID, issuing authority and scopes. Otherwise, operations will fail with opaque error messages.

I encountered this doing test development of my own, but I'm guessing this is the underlying issue here: https://github.com/iTwin/itwinjs-core/discussions/3483#discussioncomment-2774943

We should do the same for Electron @evelynpreslar-bentley

### DIFF
--- a/packages/node-cli/src/Client.ts
+++ b/packages/node-cli/src/Client.ts
@@ -5,6 +5,7 @@
 // Code based on the blog article @ https://authguidance.com
 
 import * as Http from "http";
+import * as crypto from "crypto";
 import * as open from "open";
 import { assert, AuthStatus, BeEvent, BentleyError, Logger } from "@itwin/core-bentley";
 import {
@@ -76,7 +77,7 @@ export class NodeCliAuthorizationClient implements AuthorizationClient {
   public constructor(config: NodeCliAuthorizationConfiguration) {
     this._bakedConfig = new BakedAuthorizationConfiguration(config);
 
-    const appStorageKey = `iTwinJs:${this._bakedConfig.clientId}:${this._bakedConfig.issuerUrl}`;
+    const appStorageKey = makeAppStorageKey({ ...this._bakedConfig });
     this._tokenStore = new TokenStore(appStorageKey);
   }
 
@@ -289,6 +290,16 @@ export class NodeCliAuthorizationClient implements AuthorizationClient {
     const httpServer = Http.createServer(onBrowserRequest);
     httpServer.listen(new URL(redirectUrl).port);
   }
+}
+
+/**
+ * @internal
+ */
+export function makeAppStorageKey(namedArgs: {clientId: string, issuerUrl: string, scopes: string}): string {
+  // A stored credential is only valid for a combination of the clientId, the issuing authority and the requested scopes.
+  // Incorporate all these parameters into the secure storage key in order to avoid reusing mismatched credentials.
+  const scopesHash = crypto.createHash("md5").update(namedArgs.scopes).digest("hex");
+  return `iTwinJs:${namedArgs.clientId}:${namedArgs.issuerUrl}:${scopesHash}`;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


aruniverse (approved):

evelynpreslar-bentley:
Wouldn't it be simpler to keep the storage key as it is and just compare the requested scope to the cached scope, invalidating the cache if it doesn't provide the scope set being requested?

A single `issuer` + `client-id` pair shouldn't be associated with multiple sets of scopes at once, right?
I'd expect the requested scopes to differ only when an app is updated, not often enough to warrant caching multiple tokens at once.

aruniverse:
> Wouldn't it be simpler to keep the storage key as it is and just compare the requested scope to the cached scope, invalidating the cache if it doesn't provide the scope set being requested?

the cahced scope? where is that? youd have to introspect the current token right?

mgooding:
> > Wouldn't it be simpler to keep the storage key as it is and just compare the requested scope to the cached scope, invalidating the cache if it doesn't provide the scope set being requested?
> 
> the cahced scope? where is that? youd have to introspect the current token right?

I had the same question... the cached data doesn't include the scopes. I had thought about changing the stored blob to wrap the cached data and also put the scope in there, but thought this was simpler.

evelynpreslar-bentley:
> > > Wouldn't it be simpler to keep the storage key as it is and just compare the requested scope to the cached scope, invalidating the cache if it doesn't provide the scope set being requested?
> > 
> > 
> > the cahced scope? where is that? youd have to introspect the current token right?
> 
> I had the same question... the cached data doesn't include the scopes. I had thought about changing the stored blob to wrap the cached data and also put the scope in there, but thought this was simpler.

Our browser auth client caches the scope alongside the token, but we're not comparing it to anything because those tokens expire so fast.

I'm mainly concerned about the cache becoming filled with tokens stored under scope sets that aren't used anymore after a code change.

mgooding:
> > > > Wouldn't it be simpler to keep the storage key as it is and just compare the requested scope to the cached scope, invalidating the cache if it doesn't provide the scope set being requested?
> > > 
> > > 
> > > the cahced scope? where is that? youd have to introspect the current token right?
> > 
> > 
> > I had the same question... the cached data doesn't include the scopes. I had thought about changing the stored blob to wrap the cached data and also put the scope in there, but thought this was simpler.
> 
> Our browser auth client caches the scope alongside the token, but we're not comparing it to anything because those tokens expire so fast.
> 
> I'm mainly concerned about the cache becoming filled with old tokens stored under scope sets that aren't used anymore after a code change.

Fair enough - I can make this change. I'll need to change the key slightly since I'm changing the stored data, but that's not a big deal.

evelynpreslar-bentley:
> > > > > Wouldn't it be simpler to keep the storage key as it is and just compare the requested scope to the cached scope, invalidating the cache if it doesn't provide the scope set being requested?
> > > > 
> > > > 
> > > > the cahced scope? where is that? youd have to introspect the current token right?
> > > 
> > > 
> > > I had the same question... the cached data doesn't include the scopes. I had thought about changing the stored blob to wrap the cached data and also put the scope in there, but thought this was simpler.
> > 
> > 
> > Our browser auth client caches the scope alongside the token, but we're not comparing it to anything because those tokens expire so fast.
> > I'm mainly concerned about the cache becoming filled with old tokens stored under scope sets that aren't used anymore after a code change.
> 
> Fair enough - I can make this change. I'll need to change the key slightly since I'm changing the stored data, but that's not a big deal.
> 
> While I'm at it - thoughts on moving the issuerUrl into this blob and out of the key as well? Makes the entries easier to read in Credential Manager and I don't think it's a big lift to sign in again when switching environments...

For the browser auth client, it makes sense to include it to prevent random collisions across sites that happen to share a client ID.
For the cli and electron clients, it's probably less important, but I'd still advocate to keep it unless readability is preferable.
For tokens that can silently refresh for a long time I can see how re-signing in could get annoying.

For the key change, can't you just invalidate the cache if a mismatching or `undefined` scope is read?
That way all the old scopeless values don't get reused even though the key matches.

mgooding:
@evelynpreslar-bentley @calebmshafer Please review when you get a chance.

evelynpreslar-bentley (approved):

PR #82, opened 2022-08-08
Specify supported Electron versions through peer dependency

GytisCepk:
## Changes

- Added "electron" peer dependency with range `>=14.0.0 <18.0.0` in `@itwin/electron-authorization` package (to match range in core repo). 
- Increased "electron" dev dependency to `^17.0.0` to use latest supported electron version in development.

Related: [PR #3939: Add support for Electron versions from 15 to 17](https://github.com/iTwin/itwinjs-core/pull/3939).


aruniverse (approved):
Missing rush change.

PR #85, opened 2022-09-20
Fix OIDC signin tool cache

JuliusArmalis:


paulius-valiunas (approved):

common/config/azure-pipelines/ci.yml
```diff
@@ -1,68 +0,0 @@
-parameters:
-  - name: shouldPublish
```
aruniverse, line 2: Why have the pipelines been deleted?

JuliusArmalis: Because it was duplicated inside "common" and ".github". The one used in PRs seems to be in ".github" so I removed unused files to avoid confusion in the future.

aruniverse: Im pretty sure the one in the github folder is just the ci pipeline that runs on prs on gh;
the ones in the azure-pipelines are used to actually publish the pkg and run the ci on ado.
please revert deleting these 3 files

JuliusArmalis: Reverted the deletion

aruniverse (approved):

common/config/rush/command-line.json
```diff
@@ -40,6 +40,16 @@
       "ignoreMissingScript": false,
       "ignoreDependencyOrder": true,
       "allowWarningsInSuccessfulBuild": true
+    },
+    {
+      "name": "test",
```
wgoehrig, line 45: Why was this necessary?  I thought `rush test` already worked...

JuliusArmalis: No, it was not a command that was setup with rush. In the pipeline we run rush cover instead, but we figured it would still be nice to have rush test available for local testing.

wgoehrig (approved):

PR #88, opened 2022-10-05
Update browser authorization to oidc-client-ts

evelynpreslar-bentley:


evelynpreslar-bentley:
oidc-client-ts is the currently active fork of oidc-client-js.

While the API is mostly identical, oidc-client-ts does makes [some breaking changes](https://github.com/authts/oidc-client-ts/commit/e369d1317eee3669d7d2b98986f155ad959ba50c#diff-36186878c3c97268c76a6ebf12d92327d66bc20571bfdb45c479877ffc481b36R17-R27) in its client constructor.
The client_id and authority parameters are now required, and necessary in order to process callbacks (previously, only a redirectURL was required, and the other properties were inferred).

This has the direct effect of requiring more upfront configuration in order to call `BrowserAuthorizationCallbackHandler.handleCallback`, but also has the indirect effect of disallowing callbacks to be processed with a single callback handler configuration.

Because oidc-client-ts [no longer permits](https://github.com/authts/oidc-client-ts/commit/8de0cf64a2a3d79bfc71b2dc027ba7c92708bf82#diff-3e4d50937c7b243b88ad95b27dd8560f04619a92fbe0d07ec64ee8e3eb63ccc1L94) authority and clientId to be loaded from the state during a callback, a callback handler must be created for each authority/clientId combination. Otherwise, the state will not be loaded due to mismatching configuration.

This is of concern to any non-IMS services or extensions which utilize BrowserAuthorizationClient (like the SafetiBase frontstage in Design Review), since they are reliant on the parent app to handle callbacks for them.

common/changes/@itwin/browser-authorization/evelyn.preslar-browser-auth-update-oidc-client-ts_2022-10-05-20-04.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/browser-authorization",
+      "comment": "update to oidc-client-ts",
+      "type": "major"
```
aruniverse, line 6: while i agree its a breaking change, since its 0.x its technically allowed
wonder if we should test with this new pkg first before going to 1.0

evelynpreslar-bentley: I think it would be a good move to get this in before 1.0 simply because the old library is no longer actively maintained. We could then work around the issues I described afterwards.

I don't think it would be too complicated to work around the breaking changes, though.
We could just read the state from the URL and pre-load the newly required arguments from there (effectively the same functionality as before), but some sort of whitelist might be desirable from a security perspective.

aruniverse: yea that seems ok to me to; so if you want this in before 1.0, change the change type from `major` to `minor`

mattbjordan: @aruniverse Evelyn finished this change.

ben-polinsky:
@aruniverse trying to follow this. It's definitely a breaking change - do we need to test or update anything dependent on this package?

aruniverse:
> @aruniverse trying to follow this. It's definitely a breaking change - do we need to test or update anything dependent on this package?

yea im worried what the fallout of this will look like. 

@mattbjordan can you pull this into a cospace with a viewer and see what happens?

common/config/rush/pnpm-lock.yaml
```diff
@@ -19,22 +19,22 @@ importers:
       eslint: ^7.32.0
       mocha: ^8.2.3
       nyc: ^15.1.0
-      oidc-client: ^1.11.5
+      oidc-client-ts: ^2.0.6
```
aruniverse, line 22: what happened in this lockfile?

seems like someone did an `update --full` but most of the deps look like they were downgraded

packages/browser/package.json
```diff
@@ -27,7 +27,7 @@
     "url": "https://github.com/iTwin/auth-clients"
   },
   "dependencies": {
-    "oidc-client": "^1.11.5"
+    "oidc-client-ts": "^2.0.6"
```
aruniverse, line 30: Just want to confirm this is ok, https://github.com/authts/oidc-client-ts/blob/main/README.md?plain=1#L16

@laurynasr

mattbjordan:
> @mattbjordan can you pull this into a cospace with a viewer and see what happens?

Will do. Gotta figure out how to use cospace though as I haven't needed to before.

packages/browser/package.json
```diff
@@ -27,7 +27,7 @@
     "url": "https://github.com/iTwin/auth-clients"
   },
   "dependencies": {
```
aruniverse, line 29: @mattbjordan please make sure we're using the latest version of this pkg

common/config/rush/pnpm-lock.yaml
mattbjordan: Huh weird. Looks like doing `rush update --full` correctly updates them to the latest though. I'm going to commit this in a moment.

packages/browser/package.json
mattbjordan: Done. Updated to latest version

packages/browser/src/utils.ts
```diff
@@ -0,0 +1,17 @@
+const imsUrl = "https://ims.bentley.com";
+
+/**
```
aruniverse, line 3: what happens if someone wants to us imsoidc.bentley.com as the authority?

packages/browser/src/Client.ts
```diff
@@ -77,14 +79,7 @@ export class BrowserAuthorizationClient implements AuthorizationClient {
...
     this._basicSettings = configuration;
-
-    let prefix = process.env.IMJS_URL_PREFIX;
-    const authority = new URL(this._basicSettings.authority ?? this.url);
-    if (prefix && !this._basicSettings.authority) {
-      prefix = prefix === "dev-" ? "qa-" : prefix;
-      authority.hostname = prefix + authority.hostname;
-    }
-    this.url = authority.href.replace(/\/$/, "");
+    this.authorityUrl = configuration.authority ?? getImsAuthority();
```
aruniverse, line 82: basicSettings will alredy contain the authority. why not use that? 
now you have 2 src of "truth" and theyre possibily different

packages/browser/src/CallbackHandler.ts
```diff
@@ -7,23 +7,33 @@
...
+import { getImsAuthority } from "./utils";
 
 /**
  * @beta
  */
 export interface BrowserAuthorizationCallbackHandlerConfiguration {
+  /** The URL of the OIDC/OAuth2 provider. If left undefined, the Bentley auth authority will be used by default. */
+  readonly authority?: string;
+  /** The unique client id registered through the issuing authority. Required to obtain authorization from the user. */
+  readonly clientId: string;
```
aruniverse, line 24: the addition of these now required properties is breaking. please test thoroughly

packages/browser/src/CallbackHandler.ts
```diff
@@ -44,14 +54,16 @@ export enum OidcCallbackResponseMode {
...
   protected _userManager?: UserManager;
 
   protected _basicSettings: BrowserAuthorizationCallbackHandlerConfiguration;
   protected _advancedSettings?: UserManagerSettings;
 
-  private constructor(configuration: BrowserAuthorizationCallbackHandlerConfiguration = {}) {
+  private constructor(configuration: BrowserAuthorizationCallbackHandlerConfiguration) {
     this._basicSettings = configuration;
     BrowserAuthorizationLogger.initializeLogger();
+    this.authorityUrl = configuration.authority ?? getImsAuthority();
```
aruniverse, line 66: same comment about this as before in the client

mattbjordan:
> @mattbjordan can you pull this into a cospace with a viewer and see what happens?

So, I got the cospace working and was able to run the viewer and connect to an iModel just fine. The only thing I had to change was having to use a `BrowserAuthorizationCallbackHandlerConfiguration` in the viewer.

Before
```ts
await BrowserAuthorizationCallbackHandler.handleSigninCallback(
  oidcConfiguration.redirectUri
);
```

After
```ts
await BrowserAuthorizationCallbackHandler.handleSigninCallback({
  redirectUri: oidcConfiguration.redirectUri,
  clientId: oidcConfiguration.clientId,
  authority: oidcConfiguration.authority // optional
});
```

Is there anything specific that you'd want me to test while I have this environment set up?

packages/browser/src/utils.ts
```diff
@@ -0,0 +1,17 @@
+const imsUrl = "https://ims.bentley.com";
```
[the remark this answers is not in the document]
mattbjordan: Fixed.

packages/browser/src/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+
```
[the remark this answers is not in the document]
mattbjordan: Fixed. I kept the variable outside the function as it feels correct-enough being that it is a string constant, plus there's only one function in `utils.ts` right now anyway.

packages/browser/src/utils.ts
mattbjordan: `getImsAuthority()` is used as a fallback. E.g.,
```ts
this.authorityUrl = config.authority ?? getImsAuthority();
```

To specify a different url, it can be set in the configuration that is passed in, i.e., in a `BrowserAuthorizationClientConfiguration` or `BrowserAuthorizationCallbackHandlerConfiguration`. E.g.,
```ts
await BrowserAuthorizationCallbackHandler.handleSigninCallback({
  redirectUri: oidcConfiguration.redirectUri,
  clientId: oidcConfiguration.clientId,
  authority: "imsoidc.bentley.com" // optional
});
```

packages/browser/src/Logger.ts
```diff
@@ -62,12 +62,10 @@ export class BrowserAuthorizationLogger implements IOidcClientLogger {
...
     const logLevel = BrowserAuthorizationLogger.getLogLevel(BrowserAuthorizationLoggerCategory.Authorization);
     if (!BrowserAuthorizationLogger.initialized) {
-      OidcClientLog.logger = new BrowserAuthorizationLogger();
+      OidcClientLog.setLogger(new BrowserAuthorizationLogger());
     }
 
-    if (OidcClientLog.level < logLevel) {
-      OidcClientLog.level = logLevel;
-    }
+    OidcClientLog.setLevel(logLevel);
```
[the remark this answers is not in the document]
evelynpreslar-bentley: Checking the level was removed in oidc-client-ts, and for an initialization function, I thought it made sense to just override/reset the level anyway

packages/browser/src/Client.ts
evelynpreslar-bentley: `_basicSettings` is set to `configuration` on the line just above, so they're guaranteed to be the same value here. Is there a best practice for using the member variable instead of the local one?

evelynpreslar-bentley: This should probably just be a getter that checks `_basicSettings` and `_advancedSettings`

packages/browser/src/Logger.ts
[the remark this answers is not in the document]
mattbjordan: I just pushed some "big" changes regarding this. They're not really "big" because I simply removed code that was meaningless. (I can't speak for if it was meaningless before the change from `oidc-client` to `oidc-client-ts`.) Messing with `oidc-client-ts`'s log levels is not necessary since the log levels are all managed by the `core-bentley` logger. 

The `BrowserAuthorizationLogger` class is overloading (technically implementing an interface) `oidc-client-ts`'s default logging functions to just straight up call Bentley's logging functions instead. Bentley's logger handles the log-level checking so there's no need to do any of that in `BrowserAuthorizationLogger`. This even makes everything work correctly despite `oidc-client-ts` and `core-bentley` having different log-level hierarchies.

packages/browser/src/utils.ts
aruniverse: so if i want to use qa imsoidc, i need to explicitly pass that in becasue we will only inject the env prefix for our default?

packages/browser/src/Logger.ts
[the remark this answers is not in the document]
aruniverse: > Messing with oidc-client-ts's log levels is not necessary since the log levels are all managed by the core-bentley logger.

idk if i agree with this. i cant imagine we've added exhaustive logs on our end compared to the actual lib we're using. imo we should still set the oidc clients log level

our log level enums from core-bentley dont match the levels from oidc-client-ts so i still think we need the f you just removed

packages/browser/src/Client.ts
```diff
@@ -237,7 +228,7 @@ export class BrowserAuthorizationClient implements AuthorizationClient {
 
     // Attempt a silent sign-in
     try {
-      user = await userManager.signinSilent(); // calls events
+      user = await userManager.signinSilent() ?? undefined; // calls events
```
ben-polinsky, line 231: Since `signinSilent` now returns `Promise<User|null>` you can modify the `loadUser` method to return ull instead of undefined. Then this addition here should not be necessary.

packages/browser/src/Logger.ts
```diff
@@ -7,9 +7,10 @@
  * @module Logging
  */
 
-import type { Logger as IOidcClientLogger} from "oidc-client";
-import { Log as OidcClientLog } from "oidc-client";
 import { Logger, LogLevel } from "@itwin/core-bentley";
+import type { ILogger as IOidcClientLogger } from "oidc-client-ts";
+import { Log as OidcClientLog } from "oidc-client-ts";
+// import { Logger } from "oidc-client-ts"
```
ben-polinsky, line 13: nit: remove

packages/browser/src/Logger.ts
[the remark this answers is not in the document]
evelynpreslar-bentley: > The `BrowserAuthorizationLogger` class is overloading (technically implementing an interface) `oidc-client-ts`'s default logging functions to just straight up call Bentley's logging functions instead. Bentley's logger handles the log-level checking so there's no need to do any of that in `BrowserAuthorizationLogger`. This even makes everything work correctly despite `oidc-client-ts` and `core-bentley` having different log-level hierarchies.

The logs emitted from the library are very useful for debugging, and without setting up `OidcClientLog`, those logs aren't forwarded to core-bentley's logger. It's necessary for `BrowserAuthorizationLogger` to implement the `IOidcClientLogger` interface because `OidcClientLog.setLogger` expects that type.

packages/browser/src/utils.ts
evelynpreslar-bentley: If it makes sense to, an environment variable could be used within `getImsAuthority` to change the default IMS authority url.
Not sure if there's already one for that we could adopt.
Or if imsoidc would just be a better fit for the constant value.

packages/browser/src/Client.ts
evelynpreslar-bentley: The coalescing operator here is to match the return type of `nonInteractiveSignIn`. Is `null` better than `undefined` there? I was trying to minimize usage of `null` in favor of `undefined` to match the best practices in the main itwinjs repo

packages/browser/src/Logger.ts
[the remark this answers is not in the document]
mattbjordan: Ahh, I think I see what you two are saying? That `oidc-client-ts` has its own logs that we want to keep around as well. Plus, we would need to call `OidcClientLog.setLevel` to trigger those logs to function; and also to trigger the log messages to be "forwarded" to Bentley's logs I think...

I'll have to take another shot at figuring out how I can trigger `oidc-client-ts` to log a message other than just calling `BrowserAuthorizationLogger`'s methods so that I can test this better.

packages/browser/src/Client.ts
ben-polinsky: My understanding was that `nonInteractiveSignin`'s return type matched the oidc-client's possible User types of: `User | undefined`, and that the new library uses `null` instead of `undefined` to represent a nonexistent user. 

But if the guidance is to prefer undefined over null, then I've got no problem with it. Where are these best-practices btw?

evelynpreslar-bentley: The iTwin.js coding guidelines can be found [here](https://www.itwinjs.org/v1/learning/guidelines/typescript-coding-guidelines/#do-not-use-null)

packages/browser/src/Logger.ts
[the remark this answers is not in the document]
mattbjordan: I've reverted the change I did. I spent a bit of time messing around with it by printing the logs to the console and running a viewer in a cospace and I understand what's going on a lot better.

And to wrap back around to Arun's original question, with the new package there is no way to even check the log level  without keeping track of it yourself. Since we forward the logs to Bentley's logger though, _it_ handles any sort of log level checking that is necessary. I.e., every log that `oidc-client-ts` wants to write is unconditionally forwarded to Bentley's logger and also gets level-checked by Bentley's logger.

packages/browser/src/Client.ts
mattbjordan: I did what Evelyn suggested and made a `getAuthorityUrl` method that checks for a value in `_advancedSettings` first, then `_basicSettings`. I can make the same changes to `CallbackHandler.ts` if what I did is acceptable.

packages/browser/src/CallbackHandler.ts
evelynpreslar-bentley: `oidc-client-ts` made the `client` property required to construct the `UserManager`. Instead of trying to circumvent this new requirement to keep things the same, I changed it so I could just test the library itself and fish for feedback after I confirmed that it worked.

Do you think it would be preferable to make this property optional again as a part of this PR, or in a later one?

packages/browser/src/Client.ts
```diff
@@ -65,26 +67,21 @@ export const isBrowserAuthorizationClient = (client: AuthorizationClient | undef
  */
 export class BrowserAuthorizationClient implements AuthorizationClient {
   public readonly onAccessTokenChanged = new BeEvent<(token: AccessToken) => void>();
-  public url = "https://ims.bentley.com";
   protected _userManager?: UserManager;
 
-  protected _basicSettings: BrowserAuthorizationClientConfiguration;
+  protected _basicSettings: MarkRequired<BrowserAuthorizationClientConfiguration, "authority">;
```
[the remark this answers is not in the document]
mattbjordan: @aruniverse Can you look the comments above about renaming this type / adding a new type? I wanted to ask you about this when we spoke but forgot and I didn't realize that the comments were "pending" on Github.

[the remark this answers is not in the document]
mattbjordan: This would look one of three ways. Which do you think is best? Or maybe I'm overcomplicating things and should just leave it as is.

Config is an _exported_ **interface**. Options is extended to make `authority` required in Config.
```ts
export interface BrowserAuthorizationCallbackHandlerConfiguration extends
  BrowserAuthorizationCallbackHandlerConfigurationOptions {
  /** The URL of the OIDC/OAuth2 provider. */
  readonly authority: string;  // Note: Required property
}
```

Config is an _exported_ **interface**. Use MarkRequired to make `authority` required, but convert the type into an interface.
```ts
export interface BrowserAuthorizationCallbackHandlerConfiguration extends
  MarkRequired<BrowserAuthorizationCallbackHandlerConfigurationOptions, "authority"> { }
```

Config is a _NOT exported_ **type** and is local only to CallbackHandler.ts. Use MarkRequired to make `authority` required
```ts
type BrowserAuthorizationCallbackHandlerConfiguration =
  MarkRequired<BrowserAuthorizationCallbackHandlerConfigurationOptions, "authority">;
```

I think the last option is the best because the Config is only required by the `BrowserAuthorizationCallbackHandler` class. While the Options are what an outside user would need to see.

packages/browser/src/utils.ts
mattbjordan: - There is no standardized list of environment variables; and hardcoding "https://ims.bentley.com" is how it's done in many other places in itwinjs-core.
- `getImsAuthority` exists as a function because it converts the prefix "dev-" to "qa-" by default. The simplification mentioned above wouldn't keep that functionality. See #18

packages/browser/src/CallbackHandler.ts
mattbjordan: The same changes made to Client.ts have been made in CallbackHandler.ts. (Arun's comments on Client.ts here: https://github.com/iTwin/auth-clients/pull/88#discussion_r1069861645)

packages/browser/src/utils.ts
aruniverse: ok lets leave this as is for now; i dont think we need to create an instance of Url, can just figure out the prefix as is, and just return a templated string and use the local prefix var instead of process.env.IMJS_URL_PREFIX

aruniverse:
> This is of concern to any non-IMS services or extensions which utilize BrowserAuthorizationClient (like the SafetiBase frontstage in Design Review), since they are reliant on the parent app to handle callbacks for them.

@jason-crow @jjbeckman13  fyi

packages/browser/src/Client.ts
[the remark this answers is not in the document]
aruniverse: The way as is ok to me, id just also update the type of the constructor arg to also make sure its required there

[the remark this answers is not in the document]
aruniverse: actually since BrowserAuthorizationClientConfiguration is our type, just make it required there

packages/browser/package.json
aruniverse: Confirmed it's ok

packages/browser/src/utils.ts
mattbjordan: I removed the creation of a URL object and am just returning an interpolated string now. This also lead to me removing the const default ims string. 

Also, I noticed that every package in this repo has the prefix logic code repeated (the code below). I'll make a separate issue to track making the same changes from this PR to each package. And maybe we can get other opinions on if we really should make a common function for it.
```ts
const defaultIssuerUrl = "https://ims.bentley.com";
let prefix = process.env.IMJS_URL_PREFIX;
const authority = new URL(config.issuerUrl ?? defaultIssuerUrl);
if (prefix && !config.issuerUrl) {
  prefix = prefix === "dev-" ? "qa-" : prefix;
  authority.hostname = prefix + authority.hostname;
}
this.issuerUrl = authority.href.replace(/\/$/, "");
```

aruniverse: tbh im ok with repeating this function in every pkg in this monorepo, mostly because i dont know what pkg we'd move this into, and dont want another auth-common pkg

packages/browser/src/CallbackHandler.ts
mattbjordan: @aruniverse I've finished adding some trivial tests for the constructors that at least show that the authority and other passed in values get set to the internal `_basicSettings` property correctly. 

Next week, I'll work on tests for the whole package that actually test that authorization works. We'll track that here:
[Create tests for `@itwin/browser-authorization` #101 ](https://github.com/iTwin/auth-clients/issues/101)

packages/browser/src/CallbackHandler.ts
```diff
@@ -114,25 +137,25 @@ export class BrowserAuthorizationCallbackHandler {
...
       return;
 
     let errorMessage = "";
-    let callbackHandler = new BrowserAuthorizationCallbackHandler({ responseMode: "fragment" });
+    let callbackHandler = new BrowserAuthorizationCallbackHandler({ ...config, ...{ responseMode: "fragment" } });
     try {
       await callbackHandler.handleSigninCallbackInternal();
       return;
-    } catch (err) {
+    } catch (err: any) {
```
aruniverse, line 150: we dont get anything from typing this as any

packages/browser/src/CallbackHandler.ts
```diff
@@ -114,25 +137,25 @@ export class BrowserAuthorizationCallbackHandler {
...
-  public static async handleSigninCallback(redirectUrl: string): Promise<void> {
-    const url = new URL(redirectUrl);
+  public static async handleSigninCallback(config: BrowserAuthorizationCallbackHandlerConfiguration): Promise<void> {
+    const url = new URL(config.redirectUri);
     if (url.pathname !== window.location.pathname)
       return;
 
     let errorMessage = "";
-    let callbackHandler = new BrowserAuthorizationCallbackHandler({ responseMode: "fragment" });
+    let callbackHandler = new BrowserAuthorizationCallbackHandler({ ...config, ...{ responseMode: "fragment" } });
```
aruniverse, line 146: i dont think we need to spread again for responseMode
i think this would suffice:
```ts
new BrowserAuthorizationCallbackHandler({ ...config, responseMode: "fragment" });
```

packages/browser/src/CallbackHandler.ts
```diff
@@ -46,12 +58,20 @@ export enum OidcCallbackResponseMode {
 export class BrowserAuthorizationCallbackHandler {
```
aruniverse, line 58: do we really need this separate class? i want to see if its possible to remove this altogether, and consolidate whats needed in the Client.

can handle this in a separate pr

packages/browser/src/CallbackHandler.ts
mattbjordan: Fixed

packages/browser/src/CallbackHandler.ts
ben-polinsky: Matt's right, though this is annoying in general, you need to cast to unknown and then you can attempt to cast to a known error type (or otherwise test for `.message`) - the rationale is we _think we know_  which error is being thrown, but we don't control the stack all the way so, maybe a different type of error is being thrown.

packages/browser/src/CallbackHandler.ts
mattbjordan: You might be right. We can make as separate issue to track looking into this if you want. I see that the CallbackHandler ultimately calls `userManager.signinCallback` while Client will call `userManager.signinRedirect` or `.signInSilent` or `.signinPopup`. We'd have to figure out why it was done this way. There are matching `signinRedirectCallback`, etc. methods as well.

I'll keep this in the back of my mind as I'm writing tests this week. The tests will help me get a better understanding of how the code works. Though I may not be around to work on this issue...

aruniverse: fyi @evelynpreslar-bentley

evelynpreslar-bentley: The main incentive I had for splitting these into 2 classes was to make the config as simple as possible. Before `oidc-client-ts` made the `client-id` and `authority` required for handling callbacks, this made more sense, because all you needed was a `redirect-uri`. Unless we make the effort to preserve that simplicity, though, I agree that merging the two classes into one starts to make more sense.

Keeping them separate does make it easier for apps to handle callbacks as early as possible, since less configuration is required. Previously, I saw many SPA apps using the same client used for signin also being used to handle callbacks. This meant that half the app would load before handling redirects, because certain config variables were needed, but I don’t think that’s a good enough reason on its own to advocate keeping them separate.

evelynpreslar-bentley: Regarding the many different `callback` functions in `oidc-client`: they’re not necessary to call directly, since the general case function funnels execution down the right path by reading the state in the URL to see if it was a silent, popup, or redirect signin.

packages/browser/src/CallbackHandler.ts
mattbjordan: @aruniverse are we good here?

packages/browser/src/Client.ts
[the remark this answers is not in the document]
mattbjordan: @ben-polinsky @calebmshafer Any opinions here? Now would be the time to make this type of change. I believe @aruniverse is still conflicted on this.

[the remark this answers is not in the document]
aruniverse: There isnt much difference btw using 

```ts
interface BrowserAuthorizationCallbackHandlerConfiguration extends
  MarkRequired<BrowserAuthorizationCallbackHandlerConfigurationOptions, "authority"> { }
```

and 
```ts
type BrowserAuthorizationCallbackHandlerConfiguration =
  MarkRequired<BrowserAuthorizationCallbackHandlerConfigurationOptions, "authority">;
```

i say go with 3 and use that type internally

packages/browser/src/CallbackHandler.ts
aruniverse: to be tracked in https://github.com/iTwin/auth-clients/issues/103

packages/browser/src/test/Logger.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        BentleyLogger.setLevel(BrowserAuthorizationLoggerCategory.Authorization, BentleyLogLevel.Error);
+        oidcLogger.error("error_message");
+
+        assertMessageLogged(errorLogs, "error_message");
+      });
+    });
+  });
+
+  /**
+   * The syntax `BrowserAuthorizationLogger["getLogLevel"]()` used below is to call
```
aruniverse, line 103: good stuff

packages/browser/src/test/BrowserAuthorizationClient.test.ts
```diff
@@ -3,46 +3,83 @@
...
+      assert.equal(settings.responseType, "test_responseType");
+      assert.equal(settings.noSilentSignInOnAppStartup, false);
+      assert.equal(settings.silentRedirectUri, "test_silentRedirectUri");
+    });
+
+    it("given authority is used", () => {
+      assert.equal(testClient.authorityUrl, TEST_AUTHORITY);
+    });
+
+    it("given authority is used and environment prefix is ignored", async () => {
```
aruniverse, line 58: idk if you need these authority tests, since they should technically be covered by the tests in utils

common/changes/@itwin/browser-authorization/evelyn.preslar-browser-auth-update-oidc-client-ts_2022-10-05-20-04.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/browser-authorization",
+      "comment": "update to oidc-client-ts",
```
aruniverse, line 5: could you expand on this and add the breaking changes from this pr? (the changes to the callback handler). This is what shows in the changelog file which users will be using to look at whats changed.

common/changes/@itwin/browser-authorization/evelyn.preslar-browser-auth-update-oidc-client-ts_2022-10-05-20-04.json
```diff
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/browser-authorization",
-      "comment": "update to oidc-client-ts",
+      "comment": "Update to oidc-client-ts\\nAdd new required properties to BrowserAuthorizationClientConfiguration\\nRename BrowserAuthorizationClientConfiguration to BrowserAuthorizationClientConfigurationOptions",
```
aruniverse, line 5: why are we renaming the existing interface/type? thats breaking as well for no reason

packages/browser/src/test/BrowserAuthorizationClient.test.ts
mattbjordan: I split up the tests that were previously all together and added a new one or two. These two "given authority" tests are directly from before, but don't live with the utils tests anymore. These are testing that the BrowserAuthorizationClient works as expected when using prefixes and passing in an authority to use, as opposed to just testing the getImsAuthority() function.

packages/browser/src/CallbackHandler.ts
```diff
@@ -7,23 +7,42 @@
...
+ * The internal configuration used by BrowserAuthorizationCallbackHandler.
+ */
+type BrowserAuthorizationCallbackHandlerConfiguration =
+  MarkRequired<BrowserAuthorizationCallbackHandlerConfigurationOptions, "authority">;
 
 /**
  * @beta
  */
-export interface BrowserAuthorizationCallbackHandlerConfiguration {
+export interface BrowserAuthorizationCallbackHandlerConfigurationOptions {
```
aruniverse, line 29: leave this name as is, and just name your new type with the Options suffix

mattbjordan: I was going to refute this, but I thought about it more and am fine with it. So it changed it back.

common/changes/@itwin/browser-authorization/evelyn.preslar-browser-auth-update-oidc-client-ts_2022-10-05-20-04.json
mattbjordan: I was going to refute this, but I thought about it more and am fine with it. So it changed it back.

aruniverse (approved):

PR #90, opened 2022-11-10
update to 3.3, rm use of AuthStatus error code

aruniverse:


packages/oidc-signin-tool/src/certa/certaBackend.ts
```diff
@@ -59,7 +59,9 @@ async function signin(user: TestUserCredentials, oidcConfig?: TestBrowserAuthori
...
 }
 
-registerBackendCallback(getTokenCallbackName, async (user: any, oidcConfig?: any): Promise<string> => {
-  const accessToken = await signin(user, oidcConfig);
-  return accessToken?.toString() ?? "";
-});
+registerBackendCallback(
+  getTokenCallbackName,
+  async (user: any, oidcConfig?: any): Promise<string> => {
+    return signin(user, oidcConfig);
```
aruniverse, line 65:
```suggestion
    return await signin(user, oidcConfig);
```

calebmshafer:
this looks good to me. Let's wait until the current release is through that bumps the patch version. Then we can do this and bump the minor.

calebmshafer (approved):

PR #109, opened 2023-02-09
Start supporting Electron 23, drop older Electron versions

GytisCepk:
## Changes

- Dropped Electron versions:
  - Electron 14, 15, 16, 17 - reached end-of-life.
  - Electron 22 - uses Node 16, which we want to drop.
- Added support for Electron 23. No changes are required since none of the [breaking changes](https://github.com/electron/electron/blob/main/docs/breaking-changes.md#planned-breaking-api-changes-230) should affect us.

### DIFF
--- a/common/changes/@itwin/electron-authorization/gytis-change-Electron-support-range_2023-02-09-10-56.json
+++ b/common/changes/@itwin/electron-authorization/gytis-change-Electron-support-range_2023-02-09-10-56.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@itwin/electron-authorization",
       "comment": "Drop support for Electron 14, 15, 16, 17, 22. Start supporting Electron 23.",
-      "type": "major"
+      "type": "minor"
     }
   ],
   "packageName": "@itwin/electron-authorization"
--- a/common/changes/@itwin/electron-authorization/gytis-change-Electron-support-range_2023-02-09-10-56.json
+++ b/common/changes/@itwin/electron-authorization/gytis-change-Electron-support-range_2023-02-09-10-56.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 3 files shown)


common/changes/@itwin/electron-authorization/gytis-change-Electron-support-range_2023-02-09-10-56.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/electron-authorization",
+      "comment": "Drop support for Electron 14, 15, 16, 17, 22. Start supporting Electron 23.",
+      "type": "major"
```
aruniverse, line 6: Since this PKG is 0.x let's change this minor

aruniverse (approved):

packages/electron/package.json
```diff
@@ -50,7 +50,7 @@
   },
   "peerDependencies": {
     "@itwin/core-bentley": "^3.3.0",
```
aruniverse, line 52: Since this is intended to be used with itwinjs 4, do you think this peer dep here is going to cause issues when we try pulling it into the core repo?

GytisCepk: I need to go over "_@itwin/core-bentley_" to be sure, but I think "_@itwin/core-bentley_" is a pure package.

My only concern is how `rush` will handle when peer requires `^3.3.0` core packages and workspace will contain 4.x version of the same packages.

aruniverse: I don't recall and Im not at my desk rn to check the src but if it's using for events I worry. 

I think rush install will complain on install but it might be ok. I think we have some hacks in the pnomfile.cjs to make sure it's the same ver of core Bentley as in the monorepo

ben-polinsky (approved):

Issue #111, opened 2023-02-13
Multiple packages depend on vulnerable version of `openid-client`


MindaugasButkus:
Npm audit fails if you reference `@itwin/service-authorization` (newest version 0.7.0), or `@itwin/oidc-signin-tool` (latest version 3.6.1).

Steps to reproduce:
1. `npm init`
2. `npm i @itwin/service-authorization` - reports `4 high severity vulnerabilities`
3. `npm audit` produces this output:
```
npm audit report

cacheable-request  <10.2.7
Severity: high
cacheable-request depends on http-cache-semantics, which is vulnerable to Regular Expression Denial of Service - https://github.com/advisories/GHSA-8x6c-cv3v-vp6g
No fix available
node_modules/cacheable-request
  got  8.0.0 - 12.4.1
  Depends on vulnerable versions of cacheable-request
  node_modules/got
    openid-client  1.18.0 - 4.9.1
    Depends on vulnerable versions of got
    node_modules/openid-client
      @itwin/service-authorization  *
      Depends on vulnerable versions of openid-client
      node_modules/@itwin/service-authorization

4 high severity vulnerabilities

Some issues need review, and may require choosing
a different dependency.
```

paulius-valiunas:
FYI: https://github.com/jaredwray/cacheable-request/issues/225
I suggest you ignore that alert as it will eventually be unlisted anyway.

PR #112, opened 2023-02-13
`openid-client` upgraded to version 5.4.0

MindaugasButkus:
Solves #111

Note: `openid-client` dropped support for `retry`. I see @paulius-valiunas participated in related discussion here: https://github.com/panva/node-openid-client/discussions/450

### DIFF
--- a/common/changes/@itwin/oidc-signin-tool/mindaugas-openid-client-upgrade_2023-02-13-12-46.json
+++ b/common/changes/@itwin/oidc-signin-tool/mindaugas-openid-client-upgrade_2023-02-13-12-46.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/oidc-signin-tool",
+      "comment": "`openid-client` upgraded to 5.4.0",
+      "type": "major"
+    }
+  ],
+  "packageName": "@itwin/oidc-signin-tool"
+}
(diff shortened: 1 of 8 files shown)


paulius-valiunas:
Please see the discussion I linked in #111. I still want to update to openid-client 5.x (or get rid of this dependency altogether) but first I would explore the implications of removing retries

MindaugasButkus:
Test are failing due to `id_token` appearing during exchange of `code` for `access_token`. Openid-client became very strict about id_token. It throws error if you did not request it but the server still sent it. Could this be caused by `openid` scope? `Sign in (#integration)` tests pass locally without `openid` scope. The tests start failing with the same error as in the Azure Pipeline if I add the `openid` scope.

paulius-valiunas:
> Test are failing due to `id_token` appearing during exchange of `code` for `access_token`. Openid-client became very strict about id_token. It throws error if you did not request it but the server still sent it. Could this be caused by `openid` scope? `Sign in (#integration)` tests pass locally without `openid` scope. The tests start failing with the same error as in the Azure Pipeline if I add the `openid` scope.

yes, that sounds correct. We have a bad habit of including that scope everywhere even though we don't need it. Let me try removing that from the variable group...

aruniverse:
is this pr still needed?

MindaugasButkus:
> is this pr still needed?

@aruniverse The npm audit passes now even without this PR. I'd still say that we should upgrade `openid-client` dependency. We will have to do it eventually, so why not do it now?

paulius-valiunas:
I'm hesitant to merge this because removing the retry can have bad consequences. We've seen a lot of timeouts and other network errors when calling IMS in PROD, so if we were to use the new version of this package we would have to implement our own retry logic.

aruniverse:
might be able to close this pr, with https://github.com/iTwin/auth-clients/pull/167

PR #115, opened 2023-02-28
correctly specify user redirect state

ben-polinsky:
Missed in the upgrade to `oidc-client-ts` - user-defined state must be passed as a state property to `signinRedirect` rather than as the first argument itself.


ben-polinsky:
I'll take a look at the unit tests and see if I can test this there - but I have integration tests pending on another branch which will absolutely cover this.

evelynpreslar-bentley (approved):

Issue #103, opened 2023-01-25
Consolidate `BrowserAuthorizationCallbackHandler` into `BrowserAuthorizationClient`


mattbjordan:
Get rid of [BrowserAuthorizationCallbackHandler](https://github.com/iTwin/auth-clients/blob/main/packages/browser/src/CallbackHandler.ts) and have everything be handled by [BrowserAuthorizationClient](https://github.com/iTwin/auth-clients/blob/main/packages/browser/src/Client.ts).

The BrowserAuthorizationCallbackHandler and BrowserAuthorizationClient now share many of the same options (or will once #88 is merged), so their functionality makes more sense to be combined now.

_Notes from @evelynpreslar-bentley in https://github.com/iTwin/auth-clients/pull/88#discussion_r1084372892_

> The main incentive I had for splitting these into 2 classes was to make the config as simple as possible. Before `oidc-client-ts` made the `client-id` and `authority` required for handling callbacks, this made more sense, because all you needed was a `redirect-uri`. Unless we make the effort to preserve that simplicity, though, I agree that merging the two classes into one starts to make more sense.
> 
> Keeping them separate does make it easier for apps to handle callbacks as early as possible, since less configuration is required. Previously, I saw many SPA apps using the same client used for signin also being used to handle callbacks. This meant that half the app would load before handling redirects, because certain config variables were needed, but I don’t think that’s a good enough reason on its own to advocate keeping them separate. Apps should have a special route for redirects anyway.

PR #119, opened 2023-03-03
Add browser auth integration test

ben-polinsky:
Add playwright tests for Browser based authentication, both via redirect and via popup. 

Closes #101

PR #123, opened 2023-03-07
Consolidate BrowserAuthCallbackHandler and BrowserAuthorizationClient

ben-polinsky:
Merge BrowserAuthCallbackHandler into BrowserAuthorizationClient

Fixes: #103


ben-polinsky:
I've tested this out using the integration tests in #119 and they pass, FYI

aruniverse (commented):
thx for killing off the static

aruniverse:
> I've tested this out using the integration tests in #119 and they pass, FYI

lets get that pr merged in before this

packages/browser/src/Client.ts
```diff
@@ -402,4 +357,65 @@ export class BrowserAuthorizationClient implements AuthorizationClient {
...
+   * @param redirectUrl Checked against the current window's URL. If the given redirectUrl and the window's path don't match, no attempt is made to parse the URL for a token.
+   */
+  public async handleSigninCallback(): Promise<void> {
+    const url = new URL(this._basicSettings.redirectUri);
+    if (url.pathname !== window.location.pathname)
+      return;
+
+    let errorMessage = "";
+
+    try {
```
aruniverse, line 399: these 2 try blocks will do the same thing.
previously the handler would try to check both type of response modes (query & fragment)
callbackInternal should probably take an arg for response mode, which can then be used to determine the urlSuffix and do the actual signin callback

ben-polinsky: Right, my mistake - may have been a vestige of me trying to figure out the silent callback and fiddling around. Thanks for catching this.

ben-polinsky: Okay fixed.

packages/browser/README.md
```diff
@@ -6,7 +6,53 @@ Copyright © Bentley Systems, Incorporated. All rights reserved. See LICENSE.md
...
+```
+
+Exactly how you use the `client` will depend on your specific application and workflow. Here's one common way:
+
+```typescript
+try {
+  // will attempt to sign in silently
+  // If a user can be signed in without interaction, this will complete without error.
+  // Otherwise, an error will be thrown..
+  await client.signInSilent();
```
evelynpreslar-bentley, line 32: It's not necessary to call `signInSilent` before `signInRedirect`, since the latter attempts a non-interactive signin first (and that behavior is documented on `signInRedirect`)

That behavior can be skipped with certain `prompt` arguments, but it might be valuable to simplify this example by removing the `signInSilent` and the try-catch?
As written, this code will attempt to 2 silent signins in a row, which is redundant in all cases.

ben-polinsky: Yes, thanks - that makes sense.

WillGrayMSU (approved):

aruniverse:
@WillGrayMSU did you pull this into DR and test in a CoSpace to verify this worked fine?

WillGrayMSU:
> @WillGrayMSU did you pull this into DR and test in a CoSpace to verify this worked fine?

I haven't, but I can try now

WillGrayMSU (changes requested):
types need to be exported

packages/browser/src/index.ts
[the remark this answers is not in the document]
ben-polinsky: Ah - thanks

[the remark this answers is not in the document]
ben-polinsky: there you are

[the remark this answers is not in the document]
WillGrayMSU: Perfect, thanks!

WillGrayMSU (approved):

ben-polinsky:
@aruniverse I'm unable to test the cra-template due to some annoying pre-release/peer deps constraints. Perhaps we should publish this as a pre-release or alpha to test.

aruniverse (approved):
getting a pre-release out will be annoying, since the change is marked as minor and its currently 0.x, no one should be pulling this in automatically. i guess we can test once released

aruniverse:
> > getting a pre-release out will be annoying, since the change is marked as minor and its currently 0.x, no one should be pulling this in automatically. i guess we can test once released
> 
> Maybe it should be major... it's a breaking change

according to semver, breaking changes are allowed in 0.x

ben-polinsky:
> > > getting a pre-release out will be annoying, since the change is marked as minor and its currently 0.x, no one should be pulling this in automatically. i guess we can test once released
> > 
> > 
> > Maybe it should be major... it's a breaking change
> 
> according to semver, breaking changes are allowed in 0.x

the more you know! Thanks

PR #126, opened 2023-03-15
add postInstall script hook to install playwright/chromium

aruniverse:


common/changes/@itwin/oidc-signin-tool/playwright-install_2023-03-15-14-18.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/oidc-signin-tool",
+      "comment": "",
```
aruniverse, line 5:
```suggestion
      "comment": "add postInstall script hook to install playwright/chromium",
```

common/config/azure-pipelines/integration-test.yaml
```diff
@@ -81,9 +81,8 @@ jobs:
 
   - script: node common/scripts/install-run-rush.js install
     displayName: Rush Install
-
-  - script: npx playwright install
-    displayName: Install Playwright
+    env:
+      PLAYWRIGHT_BROWSERS_PATH: $(Build.BinariesDirectory)/pw-browsers
```
aruniverse, line 85: do you think they'll need to set this env in the yaml files for the op tests?

packages/oidc-signin-tool/installPlaywright.js
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+const { exec } = require("child_process");
+
+const INSTALL_SCRIPT = "npx playwright install chromium";
+
+// const controller = new AbortController();
```
aruniverse, line 10: could you delete the commented out code?

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
```diff
@@ -512,6 +517,19 @@ export class TestBrowserAuthorizationClient implements AuthorizationClient {
...
     return page;
   }
+
+  private async cleanup(
+    page: Page,
+    signal: AbortSignal,
+    waitForCallbackUrl: Promise<Request | boolean>
+  ) {
+    if (signal.aborted)
+      await page.reload();
```
aruniverse, line 527: if aborted why do we need to reload the pg?
shouldnt it just continue to cleanup as normal?

common/config/azure-pipelines/integration-test.yaml
aruniverse: cc @paulius-valiunas

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
ben-polinsky: I deleted the comment I had... 
If we're on an error page then the waitForRequest callback waiting on the callback URL will not fire again. Without the page reload, the promise is unresolved at exit and playwright must try to resolve it. However, the page object is gone at this point (we've closed it) - it's not just the callback itself which is fired.

The reload is necessary to get the callback to fire once more and resolve the promise.

common/config/azure-pipelines/integration-test.yaml
ben-polinsky: I fear, yes...

ben-polinsky: Removed all playwright variables and I believe we are good.

ben-polinsky (approved):

PR #144, opened 2023-03-27
Update 0.12.0 of @itwin/electron-authorization

karolis-zukauskas:
### Breaking changes:
- Changes to `ElectronMainAuthorizationConfiguration`:
  - Property `scope` renamed to `scopes`
  - Property `redirectUri: string | undefined` changed to `redirectUris: string[]`
- Scope `offline_access` is no longer unconditionally added to `scopes` - uers should explicitly include all scopes they are requeting.
- Parameter `prompt=consent` is no longer unconditionally added to all OIDC Authentication requests
- Parameter `access_type=offline` is no longer added to OIDC Authentication requests (should have no effect on behavior)
- `TokenStore` is no longer exposed from `ElectronMainAuthorization`
- `ElectronRendererAuthorization` constructor now accepts `ElectronRendererAuthorizationConfiguration` with a required `clientId` parameter.

### Other changes:
- Refactored `TokenStore` to only save `refresh_token` (thats all we need to save)
- Added option to use custom `IpcSocket` implementation for Fronted/Backend Client communication
- Added option to specify additional OIDC parameters to be used for authentication requests (e.g. `prompt=consent`)
- Removed `redirect_uri` usage in token refresh request (not required by specification)
- Miscellaneous refactoring


packages/electron/src/renderer/Client.ts
```diff
@@ -4,46 +4,66 @@
...
   }
+
   public async signOut(): Promise<void> {
-    await this._api.invoke(ElectronAuthIPCChannelNames.signOut);
+    await this._ipcSocket.invoke(this._ipcChannelNames.signOut);
   }
+
   public async getAccessToken(): Promise<AccessToken> {
-    const token = await this._api.invoke(ElectronAuthIPCChannelNames.getAccessToken);
+    const token = await this._ipcSocket.invoke(this._ipcChannelNames.getAccessToken);
```
ben-polinsky, line 28: Is there a need to set to a token variable? Little cleaner to just return the invoke method result, perhaps.

ben-polinsky (approved):

packages/electron/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/electron-authorization",
-  "version": "0.12.0",
+  "version": "0.13.0",
```
aruniverse, line 3: revert;
changes are managed by `rush change` entries

common/changes/@itwin/electron-authorization/redirectUris_2023-03-27-15-08.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/electron-authorization",
+      "comment": "",
```
aruniverse, line 5: set `type` to `minor`,
and for comments, id add the breaking changes you are looking to make

packages/electron/src/main/Client.ts
```diff
@@ -25,33 +25,67 @@ import {
...
+  readonly redirectUris: string[];
 
   /** Client application's identifier as registered with the OIDC/OAuth2 provider. */
   readonly clientId: string;
 
-  /** List of space separated scopes to request access to various resources. */
-  readonly scope: string;
+  /**
+   * List of space separated scopes to request access to various resources.
+   * 'offline_access' scope must be included if application wants to receive a refresh token.
```
aruniverse, line 86: why is this change being made? (the not auto adding the `offline_access`) 
its a weird scope, because its not even exposed on developer.bentley.com, so most people won't even be aware of it

packages/electron/src/main/Client.ts
```diff
@@ -25,33 +25,67 @@ import {
...
+   * @note Upon signing in, the client application receives a response from the Bentley IMS OIDC/OAuth2 provider at given URI.
+   * For mobile/desktop applications, given redirect URIs must start with `http://localhost:${redirectPort}` or `https://localhost:${redirectPort}`.
+   * In the unlikely, but possible case of a port collision, it is recommended to use multiple (e.g. three) redirect URIs with different ports.
+   * A decent strategy for choosing ports for your application is: `3|4|5{GPR_ID}`. For example (GPR_ID used here is 1234):
+   * - `http://localhost:31234/signin-callback`
+   * - `http://localhost:41234/signin-callback`
+   * - `http://localhost:51234/signin-callback`
    */
-  readonly redirectUri?: string;
+  readonly redirectUris: string[];
```
aruniverse, line 79: is this specifically for studio? i dont understand why youd want multiple redirect uris

packages/electron/package.json
karolis-zukauskas: See https://github.com/iTwin/auth-clients/pull/145

packages/electron/src/main/Client.ts
karolis-zukauskas: John Cotter and his team are working on OIDC guidelines for desktop clients - one of them is to have backup redirect URIs in case of a PORT collision on the machine. I guess such situation is rare, but if it happens, the app won't be able to authenticate.

packages/electron/src/main/Client.ts
karolis-zukauskas: This one is weird. On one hand its a special scope and most likely every Bentley app will include it (and have it enabled on the OIDC Client web page), but on the other hand its an assumption and implicit behaviour (personally, explicit > implicit).

I'm not strongly attached to this change - if you think we should always include it implicitly, I can revert that as well.

packages/electron/src/main/Client.ts
ben-polinsky: I've had this happen to users in the past, so it's a good change.

ben-polinsky: (Different application)

PR #155, opened 2023-04-13
add static method for configless sign in callback

ben-polinsky:
After recent changes, we required an instance of a fully configured `BrowserAuthorizationClient` to call `handleSigninCallback()`. This change adds a new static method which automatically pulls client configuration from local storage. Conveniently, OIDC stores the config using the returned `state` nonce as a key, so we use that to ensure the correct configuration is being selected (consider the case of multiple clients, or possibly stale clients).

### DIFF
--- a/packages/browser/src/Client.ts
+++ b/packages/browser/src/Client.ts
@@ -424,9 +424,9 @@ export class BrowserAuthorizationClient implements AuthorizationClient {
    * from localStorage.
    */
   public static async handleSignInCallback() {
-    const StaticClient = new BrowserAuthorizationClient({} as any);
-    this.tryLoadSettingsFromStorage(StaticClient);
-    await StaticClient.handleSigninCallback();
+    const staticClient = new BrowserAuthorizationClient({} as any);
+    this.tryLoadSettingsFromStorage(staticClient);
+    await staticClient.handleSigninCallback();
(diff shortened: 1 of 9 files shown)


aruniverse (approved):

packages/browser/src/Client.ts
```diff
@@ -417,4 +417,37 @@ export class BrowserAuthorizationClient implements AuthorizationClient {
...
+
+  private static tryLoadSettingsFromStorage(
+    client: BrowserAuthorizationClient
+  ) {
+    const url = new URL(window.location.href);
+    const nonce = url.searchParams.get("state");
+
+    const storageEntry = window.localStorage.getItem(`oidc.${nonce}`);
+    if (!storageEntry)
+      throw new Error("Could not load oidc settings from local storage. Ensure the client is configured properly");
```
evelynpreslar-bentley, line 440: might be a nitpick, but don't functions with "try..." in their name usually not throw errors?

ben-polinsky: Yes - thanks. Was initially going to not throw.

ben-polinsky: done

evelynpreslar-bentley (approved):

PR #159, opened 2023-04-21
Create integration tests for Electron auth sign in and sign out

maggiemah:
Test sign-in and sign-out for electron-auth using Playwright


packages/electron/src/main/TokenStore.ts
```diff
@@ -31,8 +31,14 @@ export class RefreshTokenStore {
...
   private async getUserName(): Promise<string | undefined> {
-    if (!this._userName)
-      this._userName = await OperatingSystemUserName();
+    if (!this._userName) {
+      try {
+        this._userName = await OperatingSystemUserName();
+      }
+      catch {
+        this._userName = OperatingSystemUserName.sync();
+      }
```
ben-polinsky, line 40: Not sure about this, what is your intent?

aruniverse: in the test runs, there was an error about `OperatingSystemUserName` not being a fn.

@maggiemah can you share the actual error?

ben-polinsky:
I'd like to extract the test helpers to an internal package since they are being used by both browser and electron. It could be done in a different PR or this one depending on time.

packages/electron/src/main/TokenStore.ts
ben-polinsky: Also, this package could be upgraded to version 6 which has a slightly different API.

aruniverse: its also all esm, which might make things more difficult https://github.com/sindresorhus/username/releases/tag/v6.0.0
@ben-polinsky idk if we should upgrade or not

ben-polinsky: Ah right. Will no one think of the downstream users!

aruniverse: @maggiemah please share the error tho

maggiemah: Yes, the error was `TypeError: OperatingSystemUserName is not a function` when using the asynchronous call.

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,92 @@
...
+function getTokenStoreKey(clientId: string, issuerUrl?: string): string {
+  const authority = new URL(issuerUrl ?? "https://ims.bentley.com");
+  if (envPrefix && !issuerUrl) {
+    authority.hostname = envPrefix + authority.hostname;
+  }
+  issuerUrl = authority.href.replace(/\/$/, "");
+  return `iTwinJs:${clientId}:${issuerUrl}`;
+}
+
+test.describe.configure({ mode: 'serial', retries: 2 });
```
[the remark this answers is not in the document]
maggiemah: No they are not

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    electronApp.evaluate<string>(async ({ shell }) => {
+      return new Promise((resolve) => {
+        shell.openExternal = async (url: string) => {
+          return resolve(url);
+        };
+      });
+      // This runs in the main Electron process, parameter here is always
+      // the result of the require('electron') in the main app script.
+    }),
+    new Promise<string>((resolve) => {
```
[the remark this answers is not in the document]
maggiemah: The timeout was so that if there is no `shell.openExternal` call, `urlWhenClicked` would still be declared and the test would continue instead of waiting for the entire test timout to get the error `electronApplication.evaluate: Execution context was destroyed, most likely because of a navigation.`

packages/electron/src/main/TokenStore.ts
ben-polinsky: It's unclear to me exactly what is happening, but I've reproduced. If you require the module as opposed to importing it in TokenStore, all is well. 

Arguably, the tests should be written from the perspective of a user: the user knows nothing of the TokenStore or keytar, they would be interacting an app interface. The test app could display the user's authentication status by listening to the addAccessTokenChangeListener in the renderer. Your tests would then test that the proper status was displayed rather than prodding the TokenStore.

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    electronApp = await electron.launch({
+      args: ["./dist/integration-test/test-app/index.js"],
+    });
+    electronPage = await electronApp.firstWindow();
+  } catch (error) {
+    console.log(error);
+  }
+});
+
+test.afterAll(async () => {
```
ben-polinsky, line 45: Similar advice here, this should be afterEach

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  await page.waitForTimeout(500);
+  const accessToken = await tokenStore.load();
+  expect(accessToken).toBeDefined();
+  page.close();
+});
+
+test('sign out successful', async ({ browser }) => {
+  const page = await browser.newPage();
+  await testHelper.clickSignOut(electronPage);
+  await page.waitForTimeout(500);
```
ben-polinsky, line 88: Arbitrary waits like this are usually a last-resort. If you try the approach I suggested in the other comment about adding auth state to the test app, you can wait until that element appears using Playwright https://playwright.dev/docs/writing-tests#interactions

packages/electron/src/main/TokenStore.ts
ben-polinsky: What was the resolution here?

packages/electron/playwright.config.ts
```diff
@@ -0,0 +1,15 @@
+import { PlaywrightTestConfig } from '@playwright/test'
+
+const config: PlaywrightTestConfig = {
```
aruniverse, line 3: are the tests being run in parallel or serial?

packages/electron/src/integration-test/helpers/TestHelper.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  public async clickSignOut(electronPage: Page) {
+    await electronPage.waitForSelector("button#signOut");
+    const button = electronPage.getByTestId("signOut");
+    await button.click();
+    console.log("clicked sign out");
+  }
+
+  public async signIn(page: Page, url: string) {
+    expect(url).toBeDefined();
```
aruniverse, line 28: if youre passing a defined string as an arg, wouldnt this always be true?
is this to check the empty string case?

packages/electron/src/integration-test/helpers/loadConfig.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+// eslint@typescript-eslint/naming-convention
```
aruniverse, line 6: i dont think this comment is right

packages/electron/playwright.config.ts
```diff
@@ -0,0 +1,15 @@
+import { PlaywrightTestConfig } from '@playwright/test'
```
aruniverse, line 1: missing copyright

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,85 @@
+import { ElectronApplication, Page, expect, test } from "@playwright/test";
```
aruniverse, line 1: missing copyright

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,85 @@
...
+import { _electron as electron } from 'playwright';
+import { RefreshTokenStore } from "../main/TokenStore";
+
+const { clientId, envPrefix, email, password } = loadConfig();
+
+const signInOptions: SignInOptions = {
+  clientId,
+  email,
+  password,
+  envPrefix: envPrefix || "",
```
aruniverse, line 14: doesnt envPrfix default to "" in loadConfig?

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,85 @@
...
+function getTokenStoreKey(clientId: string, issuerUrl?: string): string {
+  const authority = new URL(issuerUrl ?? "https://ims.bentley.com");
+  if (envPrefix && !issuerUrl) {
+    authority.hostname = envPrefix + authority.hostname;
+  }
+  issuerUrl = authority.href.replace(/\/$/, "");
+  return `iTwinJs:${clientId}:${issuerUrl}`;
+}
+
+test.describe.configure({ mode: 'serial' });
```
aruniverse, line 31: can this be done at the top level? in the playwright config? like mentioned in my other comment

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  await testHelper.signIn(page, await urlWhenClicked);
+  await page.waitForLoadState('networkidle');
+  const accessToken = await tokenStore.load();
+  expect(accessToken).toBeDefined();
+  page.close();
+});
+
+test('sign out successful', async ({ browser }) => {
+  const page = await browser.newPage();
+  await testHelper.clickSignOut(electronPage);
```
aruniverse, line 80: shouldnt you be signed in before signing out? 
once you change the beforeAll to beforeEach this will be different

packages/electron/src/integration-test/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface SignInOptions {
```
aruniverse, line 1: copyright

aruniverse (commented):
please review comments/suggestions made by ben

packages/electron/src/main/TokenStore.ts
maggiemah: I used the listener to check the sign-in status, but I'm still worried about the issue of the tests not going through the sign-in process if credentials are stored. The test now only uses TokenStore to clear the stored passwords at the beginning. Is there another way to clear the passwords before running the tests?

packages/electron/src/integration-test/integration.test.ts
maggiemah: removed

packages/electron/playwright.config.ts
maggiemah: tests are serial, now by the line `workers: 1,`

packages/electron/src/integration-test/helpers/TestHelper.ts
maggiemah: This line was originally used to check if the url promise was void, so it isn't needed anymore.

packages/electron/src/integration-test/helpers/loadConfig.ts
maggiemah: removed

packages/electron/playwright.config.ts
maggiemah: added

packages/electron/src/integration-test/integration.test.ts
maggiemah: added

packages/electron/src/integration-test/integration.test.ts
maggiemah: envPrefix is `process.env.IMJS_URL_PREFIX || ""`. The env file I've been testing with has `IMJS_URL_PREFIX = "qa-"`.

packages/electron/src/integration-test/integration.test.ts
maggiemah: yes

packages/electron/src/integration-test/integration.test.ts
maggiemah: The Electron app stays signed in after the sign-in test. If I sign in again in this test, the status is still signed in even after signing out.

packages/electron/src/integration-test/types.ts
maggiemah: added

packages/electron/src/main/TokenStore.ts
ben-polinsky: Okay so, simplest way to proceed, I think, is to just require the tokenStore instead of importing in your test file. Then you should be able to remove the try/catch around getUserName. This isn't ideal, but it's better than modifying the library code to for a issue in tests. 

Then you can continue deleting tokens before each test, and the user should not be signed in prior to any actions.

packages/electron/src/integration-test/integration.test.ts
ben-polinsky: Since you're deleting the token before each test, you should need to sign in before signing out.

packages/electron/src/integration-test/integration.test.ts
aruniverse: yea exactly, so you already default to an empty string in loadConfig, so makes no point to default again in here. 
its redundant.

ben-polinsky (approved):

PR #164, opened 2023-05-02
bump @itwin deps and move from peer; browser 1.0

ben-polinsky:
Since we export types which contain types from @itwn/core-common, it's not only a dev dependency. By this logic, then, @itwin/core-bentley should also be a "standard" dependency.  

Set as major change for browser-authorization as 1.0 release is imminent.
For other packages, consider this a minor change?

### DIFF
--- a/packages/electron/package.json
+++ b/packages/electron/package.json
@@ -25,14 +25,14 @@
     "directory": "packages/electron"
   },
   "dependencies": {
+    "@itwin/core-bentley": "3.0.0 - 4",
+    "@itwin/core-common": "3.0.0 - 4",
     "@openid/appauth": "^1.3.1",
     "keytar": "^7.8.0",
     "username": "^5.1.0"
   },
   "devDependencies": {
-    "@itwin/build-tools": "^4.0.0-dev.63",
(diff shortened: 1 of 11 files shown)


aruniverse (approved):

packages/electron/package.json
```diff
@@ -61,7 +61,7 @@
     "typescript": "~4.3.5"
   },
   "peerDependencies": {
-    "@itwin/core-bentley": "^3.3.0",
+    "@itwin/core-bentley": ">=3.3.0",
```
paulius-valiunas, line 64: This is dangerous. I know we want to support both 3.x and 4.x, but what if 5.x comes out and breaks compatibility? I would prefer `^3.3.0 || ^4.0.0`

ben-polinsky: Safer. I'll get in another PR  - we have time before 5 :) Thanks.

paulius-valiunas: the problem is if someone decides to use this version of auth-clients with 5.0 in a year or two :D I mean it's kind of their fault at that point but I would prefer npm to complain about mismatched dependencies than pretend like it's fine

PR #172, opened 2023-05-12
remove AzureAD and Authing support

paulius-valiunas:
Remove Authing and Azure AD support as it is no longer actively used.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -4,7 +4,6 @@
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
   "configurations": [
-
     {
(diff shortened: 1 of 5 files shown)


ben-polinsky:
I agree... wanted to rip this out earlier but received mixed signals :) Hopefully this is fine with @calebmshafer

calebmshafer:
> I agree... wanted to rip this out earlier but received mixed signals :) Hopefully this is fine with @calebmshafer

fine by me to do it now. Don't think there's any reason this tool needs to be a generalized headless sign-in tool of any OIDC config. We can just limit it directly to the IMS implementation used by Bentley in general.

ben-polinsky (approved):

PR #178, opened 2023-06-02
pass metadata to oidc-client-ts

paulius-valiunas:
Without this metadata, the client uses `fetch` to get issuer metadata, which is not always available (especially in older Node.js versions)

### DIFF
--- a/common/changes/@itwin/oidc-signin-tool/no-fetch_2023-06-02-09-03.json
+++ b/common/changes/@itwin/oidc-signin-tool/no-fetch_2023-06-02-09-03.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/oidc-signin-tool",
+      "comment": "Fix issue with OIDC signin tool not working without fetch API",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/oidc-signin-tool"
+}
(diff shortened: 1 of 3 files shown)


paulius-valiunas:
@aruniverse I see you added usage of AbortController here:
https://github.com/iTwin/auth-clients/pull/126/files#diff-043b55e385b562642d1fdd78c8597fa90d2a309cd0497a31d68f6ce57bf4b722R159
Given that this package is supposed to support iTwin.js 3.3.0, I think it should also support Node.js 12.x, which doesn't have that API. Do you think you could make it work without the new API?

ben-polinsky (approved):

aruniverse:
> @aruniverse I see you added usage of AbortController here: [#126 (files)](https://github.com/iTwin/auth-clients/pull/126/files#diff-043b55e385b562642d1fdd78c8597fa90d2a309cd0497a31d68f6ce57bf4b722R159) Given that this package is supposed to support iTwin.js 3.3.0, I think it should also support Node.js 12.x, which doesn't have that API. Do you think you could make it work without the new API?

Ah good catch. Technically youre right, but considering Node 12 has been EOL for over a year, Im tempted to not fix this, and only do so if someone brings this up

aruniverse (approved):

Issue #150, opened 2023-04-04
Replace Rush 


ben-polinsky:
Rush is starting to cause more issues than it is worth:

- Need to break up publishing flow into versioning and publishing, but:

- Cannot customize tag names used when publishing.

Let's look into Beachball

ben-polinsky:
Cannot publish a subset of packages... https://github.com/microsoft/rushstack/issues/2041

paulius-valiunas:
https://github.com/iTwin/itwinjs-internal/issues/75 should help

PR #181, opened 2023-06-09
@itwin/oidc-signin-tool_v4.0.1

ben-polinsky:
Rush and our desired release workflow do not fully work as intended. Until we replace rush #150 this cleanup is necessary.


aruniverse (approved):

aruniverse:
prob gonna need rush change with an empty change comment and none as type to get this through

ben-polinsky:
> prob gonna need rush change with an empty change comment and none as type to get this through

Yep just pushed

PR #182, opened 2023-06-29
Remove installScript.js file and upgrade playwright to 1.35.1

anmolshres98:
In `oidc-signin-client`, removed `installScript.js` file with a postInstall script of `npx playwright install chromium` instead. We cannot entirely replace `@playwright/test` with `playwright` since we are using the test framework and just the `playwright` package does not come with test assertions at all. [See here](https://playwright.dev/docs/library#key-differences) for more details on this. We also cannot have both `playwright` and `@playwright/test` in the dependency because playwright refuses to run its test framework if the playwright package is also part of the dependencies. For similar [change](https://github.com/iTwin/itwinjs-core/pull/5681) in core repo, this was not an issue because that package wasn't using playwright's test framework. That is not the case for this repo.

Because of these reasons, I believe this is the best compromise we can get when it comes to leveraging playwright's test framework, having the chromium browser installed, and also not having potentially unsafe postInstall code execution.


common/changes/@itwin/oidc-signin-tool/as-rm-install-script_2023-06-29-17-01.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/oidc-signin-tool",
+      "comment": "Upgrade @playwright/test to 1.35.1 and remove installScript.js file",
+      "type": "none"
```
aruniverse, line 6:
```suggestion
      "type": "minor"
```

anmolshres98: for all other packages too or just for `oidc-signin-tool`?

aruniverse: just this one for now

aruniverse: actually for other pkgs, maybe set it to patch

aruniverse (approved):

PR #183, opened 2023-07-19
fix version ranges

paulius-valiunas:
we shot ourselves in the leg by using `>=` for dependencies. Once we release iTwin.js 5.0, we'll be in deep trouble if anyone is still using the current versions of auth clients. Since we don't update them often, that might be very likely. I decided to just do a separate patch release for this, to ensure that as many people as possible get it. Including it in an unrelated minor/major release would mean that people using `~` version ranges to pull in auth clients wouldn't get the update.


aruniverse (approved):

ben-polinsky (approved):
Thanks for cleaning this up.

PR #184, opened 2023-07-19
use got@12

paulius-valiunas:


ben-polinsky (approved):

packages/service/src/ServiceAuthorizationClient.ts
```diff
@@ -8,8 +8,9 @@
 
 import type { AuthorizationClient } from "@itwin/core-common";
 import type { ServiceAuthorizationClientConfiguration } from "./ServiceAuthorizationClientConfiguration";
-import type { Options as GotOptions } from "got";
-import got from "got";
+// @ts-expect-error -- this is a type-only import, it won't break the resulting .js file.
+// It also won't show up in the resulting .d.ts file because we don't re-export this type.
```
aruniverse, line 12: thx for the comment, i was going to ask what it looks like in the js file since import assertions are new and i dont think is apart of the ecmascript standard yet

aruniverse (approved):

PR #187, opened 2023-07-24
increase service authorization client timeout

paulius-valiunas:


ben-polinsky:
Should this be exposed to the user as an option?

aruniverse (approved):

PR #188, opened 2023-07-25
Cleanup after publishing v1.0.1 of `@itwin/service-authorization`

MindaugasButkus:


aruniverse:
please do the same for the other pkgs so it doesnt fail for those the next time

PR #193, opened 2023-08-04
Move away from node-keytar to electron's SafeStorage API

### LINKED ISSUE #161: Replace Keytar (Electron)
[Keytar](https://github.com/atom/node-keytar) has been deprecated to the dismay of everyone in the universe. We need to identify a replacement or fork.

hl662:
Closes #161

`node-keytar` has been deprecated, thus requiring us to figure out a way to do OS-based credentials. Electron's [SafeStorage API](https://www.electronjs.org/docs/latest/api/safe-storage) uses Chromium's `oscrypt` under the hood to encrypt and decrypt strings using the client OS's specific credential managers. That left us with just a persistent storage solution to look for, which [electron-store](https://github.com/sindresorhus/electron-store#readme) fulfills.

Moving straight to SafeStorage and removing `node-keytar` entirely would essentially clear out a user's sessions, and require them to perform single sign on again. Instead, we can slowly migrate to SafeStorage, by moving the tokens used in `node-keytar` over to `electron-store` the first time the auth client retrieves their refreshToken. Thus, no side-effects will happen, giving clients time to upgrade their auth-clients modules without hassle.


packages/electron/package.json
```diff
@@ -32,10 +32,12 @@
     "@itwin/core-common": "^3.3.0 || ^4.0.0",
     "@openid/appauth": "^1.3.1",
     "keytar": "^7.8.0",
-    "username": "^5.1.0"
+    "username": "^5.1.0",
+    "electron": "^24.0.0",
```
ben-polinsky, line 36: Any reason this was moved from peer dependencies? Either way, if we're requiring users to possibly upgrade from electron 23, then this is probably a breaking change.

aruniverse:
fyi @MichaelBelousov @johnnyd710 @GintV can you guys take a look to make sure we dont break you

packages/electron/src/test/MainAuthorizationClient.test.ts
```diff
@@ -71,9 +71,12 @@ describe("ElectronMainAuthorization Token Logic", () => {
     };
     const mockTokenResponse = new TokenResponse(mockTokenResponseJson);
 
+    const refreshToken = "old refresh token";
+    sinon.stub(RefreshTokenStore.prototype, "encryptRefreshToken" as any).returns(Buffer.from(refreshToken));
```
ben-polinsky, line 75: It's good practice to mock/stub "as close to the edge" of an app as possible. If you can, can you look into mocking the `safeStorage` methods instead of our `TokenStore`?

MichaelBelousov:
> fyi @MichaelBelousov @johnnyd710 @karolis-zukauskas @GintV can you guys take a look to make sure we dont break you

we actually just found that a PR we've tried to get in but is hanging in CI is likely caused by keytar, we're going to test if this patch fixes it, and will assess whether this breaks us. In our case, it's OK if it causes the user to log in again.

cc @johnnyd710

johnnyd710:
I quickly tried patching this PR into our application, but I get (replacing the name of our app with "appname": `no such file or direcory, mkdir C:\Users\John.DiMatteo\AppData\Roaming\@appname\app-nodejs\Config\iTwinJs:app-name:https:`. Btw, we override the electron userdata directory shortly after our app starts up to `C:\Users\John.DiMatteo\AppData\Roaming\Bentley\appname`.

Also, like Mike said, we are having issues with `keytar` in our CI, so it would be great if we could get rid of it completely. I see you still use it to migrate the old keys over. Could you make that optional?

ben-polinsky:
> I quickly tried patching this PR into our application, but I get (replacing the name of our app with "appname"):
> 
> `no such file or direcory, mkdir C:\Users\John.DiMatteo\AppData\Roaming\@appname\app-nodejs\Config\iTwinJs:app-name:https:`.
> 
> Btw, we override the electron userdata directory shortly after our app starts up to `C:\Users\John.DiMatteo\AppData\Roaming\Bentley\appname`.
> 
> Also, like Mike said, we are having issues with `keytar` in our CI, so it would be great if we could get rid of it completely. I see you still use it to migrate the old keys over. Could you make that optional?

@hl662 could you take a look at the first issue and also the failing integration tests? 

@johnnyd710 I see. It's going to be tough to do in a non-breaking (semver) way. Wondering if it's worth it for us to keep the migration. Was your build previously passing with keytar installed?

hl662:
> 

Hmm yeah, that error was also popping up in the CI build for windows. It might be an issue with the filename being too long, thus windows is splitting the filename in half - one part is a folder and the other part is the actual .json containing the  credentials. I can shorten the file name. As for dropping node-keytar, we could... Just that the side effect of this would be any app that bumps `itwin/electron-authorization` will just have to sign in again.

hl662:
> > 
> 
> Hmm yeah, that error was also popping up in the CI build for windows. It might be an issue with the filename being too long, thus windows is splitting the filename in half - one part is a folder and the other part is the actual .json containing the credentials. I can shorten the file name. As for dropping node-keytar, we could... Just that the side effect of this would be any app that bumps `itwin/electron-authorization` will just have to sign in again.

scratch what I said above, it failed because with the .json filename, part of it has the issuer_url which has `http:/` - the `/` tells all operating systems to create a folder with one half of the filename, and the other half is the actual file. This side effect is fine on Linux and Mac, but on Windows, it fails cause windows paths use `\`. Still gonna shorten the file name, probably gonna cut the issuer_url from it.

ben-polinsky:
> > 
> 
>  As for dropping node-keytar, we could... Just that the side effect of this would be any app that bumps `itwin/electron-authorization` will just have to sign in again.

This is true. I really liked the idea users wouldn't have to, though.

MichaelBelousov:
> This is true. I really liked the idea users wouldn't have to, though.


agreed. We were hoping more for a runtime option to not access keytar at all. We might have to patch one in ourselves if there is none, because it seems to hang in our CI builds and we haven't determined why yet. Ideally we'd reproduce the issue in a Linux VM and figure out why but haven't been able to yet and for us it'll be faster potentially to just stub out keytar. Up to you if you want to expose configuration for us to avoid it.

ben-polinsky:
> > This is true. I really liked the idea users wouldn't have to, though.
> 
> agreed. We were hoping more for a runtime option to not access keytar at all. We might have to patch one in ourselves if there is none, because it seems to hang in our CI builds and we haven't determined why yet. Ideally we'd reproduce the issue in a Linux VM and figure out why but haven't been able to yet and for us it'll be faster potentially to just stub out keytar. Up to you if you want to expose configuration for us to avoid it.

Ah, so your CI is failing at runtime? I had assumed during install/build. If that's the case, we can attempt to conditionally import based on an environment variable.

MichaelBelousov:
> Ah, so your CI is failing at runtime? I had assumed during install/build. If that's the case, we can attempt to conditionally import based on an environment variable.

that would be fine for us. env var or some sort of configuration API would work for us.

packages/electron/src/test/MainAuthorizationClient.test.ts
hl662: I tried to, but electron's safeStorage is an interface that contains the functions, not a class where we can mock it's prototype's methods. That's why I had to wrap the use of safeStorage inside of private methods of `TokenStore`. Maybe there's a way to do it, but I haven't found any

packages/electron/package.json
hl662: when kept as a peer dependency, the build for electron-auth would fail, so I had to move it away. I reverted my change that asked for electron 24 back to being a range

packages/electron/src/test/MainAuthorizationClient.test.ts
ben-polinsky: Okay, good enough for now, then. Thanks

hl662:
@MichaelBelousov @johnnyd710 can you try patching this PR in again? In your `.env` you can add `DISABLE_KEYTAR=True`, let's see if that helps avoiding the runtime errors

packages/electron/package.json
aruniverse: this should def be reverted; what was the build error?

aruniverse:
> > Ah, so your CI is failing at runtime? I had assumed during install/build. If that's the case, we can attempt to conditionally import based on an environment variable.
> 
> that would be fine for us. env var or some sort of configuration API would work for us.

we couldnt just wrap the dynamic import in a try catch?

packages/electron/package.json
```diff
@@ -32,10 +32,12 @@
     "@itwin/core-common": "^3.3.0 || ^4.0.0",
     "@openid/appauth": "^1.3.1",
     "keytar": "^7.8.0",
```
aruniverse, line 34: have we looked at optional dependencies?

johnnyd710:
> we couldnt just wrap the dynamic import in a try catch?

unlikely, when our CI machine uses `keytar` it doesn't throw an error - it crashes our backend

packages/electron/package.json
ben-polinsky: Briefly considered, but imagined that would trip up existing users who have it installed. Now that I read the docs, seems like it would work with `npm install --omit=optional` - then the try/catch of course.

MichaelBelousov:
> > we couldnt just wrap the dynamic import in a try catch?
> 
> unlikely, when our CI machine uses `keytar` it doesn't throw an error - it crashes our backend

But if we don't peer-install the optional dependency then the `try/catch` will just never load it and the package will continue on. So this is a viable route for us, pnpm should isolate the packages so that we never load keytar even if we have another dependency on it elsewhere (that is unused)

packages/electron/package.json
hl662: It acted up saying the electron module was not found, but the error is gone... the peerDependencies might not have been installed automatically the time I ran it. I'm moving this back to peerDependencies

.npmrc
```diff
@@ -1,2 +1,3 @@
 shared-workspace-lockfile=false
-shamefully-hoist=true
\ No newline at end of file
+shamefully-hoist=true
+auto-install-peers=true
```
MichaelBelousov, line 3: I would warn against `auto-install-peers=true`, in the studio case, our consumers (that weren't using this feature, which was none of them) were hit by unresolved transitive peer dependencies that we hadn't satisfied

packages/electron/package.json
hl662: Kept `electron` in peerDep and then added electron to dev dependency and everything is good now

.npmrc
hl662: removed from .npmrc

packages/electron/src/test/MainAuthorizationClient.test.ts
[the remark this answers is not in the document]
hl662: the 3 unit tests have been added.

packages/electron/package.json
```diff
@@ -32,10 +32,11 @@
...
-    "username": "^5.1.0"
+    "username": "^5.1.0",
+    "electron-store": "~8.1.0"
```
aruniverse, line 36: alphabetize deps

johnnyd710:
FYI, I was able to get this working with our app - it just required us to polyfill some electron utilities (like `safeStorage`) since our authentication happens on a separate node process that doesn't have access to electron, but that's our problem, not yours.

packages/electron/package.json
```diff
@@ -31,23 +31,23 @@
   "dependencies": {
     "@itwin/core-common": "^3.3.0 || ^4.0.0",
     "@openid/appauth": "^1.3.1",
-    "keytar": "^7.8.0",
+    "electron-store": "~8.1.0",
```
aruniverse, line 34: any reason for locking it down with `~` ?

packages/electron/src/integration-test/integration.test.ts
```diff
@@ -22,15 +22,19 @@ const signInOptions: SignInOptions = {
 let electronApp: ElectronApplication;
 let electronPage: Page;
 const testHelper = new TestHelper(signInOptions);
-const tokenStore = new RefreshTokenStore(getTokenStoreKey(clientId));
+const tokenStore = new RefreshTokenStore(getTokenStoreFileName(clientId),getTokenStoreKey(clientId));
 
 function getTokenStoreKey(_clientId: string, issuerUrl?: string): string {
```
aruniverse, line 27: kinda weird for clientId to be prefixed with an underscore; typically thats added to appease the linter that you have an used var

packages/electron/src/main/TokenStore.ts
```diff
@@ -5,8 +5,9 @@
 // Code based on the blog article @ https://authguidance.com
 
 import * as OperatingSystemUserName from "username";
-import { deletePassword, getPassword, setPassword } from "keytar";
-
+import { safeStorage } from "electron";
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const Store = require("electron-store"); // eslint-disable-line @typescript-eslint/no-var-requires
```
aruniverse, line 10: why cant import?

hl662:
`electron-store` creates the storage file in different locations, depending on whether the `Store` was initialized in the main Electron process or the renderer process - see explanation [here](https://github.com/sindresorhus/electron-store/issues/181). This explains why the integration tests were hanging - the `RefreshTokenStore` in the test file and the one from `ElectronMainAuthorization` pointed to different places. Solution was to manually resolve the path, depending on the OS.

packages/electron/src/main/TokenStore.ts
hl662: When I tried using `import { ElectronStore} from "electron-store"` the linter threw an error about enabling esModuleInterop in order to do so - it suggested using `import { ElectronStore } from "electron-store"` but then a syntaxError was thrown, saying ``import Store = require('electron-store')` is not supported by @babel/plugin-transform-typescript`

packages/electron/src/integration-test/integration.test.ts
hl662: Ah, let me refactor the code to make it use the `const` clientId defined at the top

packages/electron/src/main/TokenStore.ts
aruniverse: wait so the esModuleInterop didnt work?

packages/electron/src/main/TokenStore.ts
```diff
@@ -22,8 +23,17 @@ export class RefreshTokenStore {
...
   private _userName?: string;
 
-  public constructor(appStorageKey: string) {
-    this._appStorageKey = appStorageKey;
+  private _store: typeof Store;
+
+  public constructor(configFileName: string, appStorageKey: string, dir?: string) {
+    this._appStorageKey = appStorageKey
+      .replace(/[.]/g, "%2E") // Replace all '.' with UTF-8 representation
+      .replace(/[\/]/g, "%2F"); // Replace all '.' with UTF-8 representation
```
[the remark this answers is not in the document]
hl662: fixed

packages/electron/src/main/TokenStore.ts
hl662: No, it still didn't remove this error: 
`'ElectronStore' can only be imported by using 'import ElectronStore = require("electron-store")' or a default import.ts(2616)`

packages/electron/src/main/TokenStore.ts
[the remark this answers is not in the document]
MichaelBelousov: tbh  it's not "utf8 representation", it's url/percent encoding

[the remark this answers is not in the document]
MichaelBelousov: sorry @aruniverse should have made a separate comment but I already commented here before my page noticed you resolved; unresolving

hl662:
@MichaelBelousov @johnnyd710 can you review? If your app overriding the electron userData folder causes the client to behave wonky, I could add a config option to overwrite the tokenStore location for electronMainAuth

packages/electron/src/main/TokenStore.ts
aruniverse: `import * as ElectronStore  from "electron-store"` ?

hl662: it threw an error at the initialization of `this._store`: 
> Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.

hl662:
> > @MichaelBelousov @johnnyd710 can you review? If your app overriding the electron userData folder causes the client to behave wonky, I could add a config option to overwrite the tokenStore location for electronMainAuth
> 
> Yeah, it's a bit weird that there is a difference between the location we store config and settings, and the location you use for the token store. If you could provide us an option to override the tokenStore, that would be ideal.

I added a config option to ElectronMainAuthorizationConfiguration

johnnyd710 (approved):
tested and working with our app (with the node process patches/polyfills I mentioned before, since we don't have electron)

packages/electron/src/test/MainAuthorizationClient.test.ts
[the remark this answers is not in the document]
ben-polinsky: Thanks!

hl662:
> > > > @MichaelBelousov @johnnyd710 can you review? If your app overriding the electron userData folder causes the client to behave wonky, I could add a config option to overwrite the tokenStore location for electronMainAuth
> > > 
> > > 
> > > Yeah, it's a bit weird that there is a difference between the location we store config and settings, and the location you use for the token store. If you could provide us an option to override the tokenStore, that would be ideal.
> > 
> > 
> > I added a config option to ElectronMainAuthorizationConfiguration
> 
> Could you add this to the readme.md for the electron package?

Added, also updated the README to remove references to keytar. safeStorage does want to use lib-secret and gnome-keyring on linux [as discussed here](https://github.com/electron/electron/issues/32598), so I just replaced keytar with safeStorage

PR #198, opened 2023-08-21
Move node-cli from keytar to crypto and node-persist for refresh tokens

### LINKED ISSUE #192: Replace Keytar (node-cli)
Options for replacing keytar for non-electron applications are less clear.

hl662:
- use node-persist for persistent storage of refresh token
- use Node's built in crypto module to encrypt and decrypt refresh tokens stored
- add option to configure refresh token storage path
- add unit tests for the token store
Closes #192


packages/node-cli/src/TokenStore.ts
```diff
@@ -3,9 +3,11 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
-import { getPassword, setPassword } from "keytar";
 import * as OperatingSystemUserName from "username";
 import { TokenResponse } from "@openid/appauth";
+import { createCipheriv, createDecipheriv, randomBytes, scryptSync } from "crypto";
```
aruniverse, line 8:
```suggestion
import { createCipheriv, createDecipheriv, randomBytes, scryptSync } from "node:crypto";
```
for consistency

packages/node-cli/src/TokenStore.ts
```diff
@@ -14,37 +16,84 @@ import { TokenResponse } from "@openid/appauth";
...
-
-  public constructor(namedArgs: {clientId: string, issuerUrl: string, scopes: string}) {
+  private readonly _store: NodePersist.LocalStorage;
+  public constructor(namedArgs: {clientId: string, issuerUrl: string, scopes: string}, dir?: string) {
     // A stored credential is only valid for a combination of the clientId, the issuing authority and the requested scopes.
     // We make the storage key a combination of clientId and issuing authority so that keys can stay cached when switching
     // between PROD and QA environments.
     // We store the scopes in our password blob so we know if a new token is required due to updated scopes.
-    this._appStorageKey = `iTwinJs:${namedArgs.clientId}:${namedArgs.issuerUrl}`;
+    const configFileName = `itwinJs_${namedArgs.clientId}`;
```
aruniverse, line 25:
```suggestion
    const configFileName = `iTwinJs_${namedArgs.clientId}`;
```

is this the same structure as the electron auth token?

common/changes/@itwin/node-cli-authorization/nam-node-cli-keytar_2023-08-21-15-16.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/node-cli-authorization",
+      "comment": "Move node-cli from keytar to use crypto and node-persist",
+      "type": "minor"
```
aruniverse, line 6: well this is breaking

aruniverse (commented):
how do we handle when new scopes are added to a client and they need to resign in?

hl662:
> how do we handle when new scopes are added to a client and they need to resign in?

The TokenStore compares the new scopes with the scopes in the cached tokenResponse. If different, the load() returns undefined, prompting the user to sign in again. For extra measures, I'll delete the cache entry as well if scopes are mismatched, in case user doesn't go through the resign process (also added a unit test to confirm this behavior)

common/changes/@itwin/node-cli-authorization/nam-node-cli-keytar_2023-08-21-15-16.json
hl662: changed type to `major`

common/changes/@itwin/node-cli-authorization/nam-node-cli-keytar_2023-08-21-15-16.json
aruniverse, line 1: can delete this file

hl662: done

ben-polinsky (approved):

packages/node-cli/src/TokenStore.ts
```diff
@@ -14,39 +16,88 @@ import { TokenResponse } from "@openid/appauth";
...
+   * @returns
+   */
+  private generateCipherKey(): Buffer {
+    return scryptSync(this._appStorageKey, "iTwin", 32); // aes-256-cbc requires a key length of 32 bytes.
+  }
+  /**
+   * Uses node's native `crypto` module to encrypt the given cache entry.
+   * @returns an object containing a hexadecimal encoded token, returned as a string, as well as the initialization vector.
+   */
+  private encryptCache(cacheEntry: object): {encryptedCache: string, iv: string} {
```
ben-polinsky, line 61: Can this cacheEntry be better typed?

GintV:
> > how do we handle when new scopes are added to a client and they need to resign in?
> 
> The TokenStore compares the new scopes with the scopes in the cached tokenResponse. If different, the load() returns undefined, prompting the user to sign in again. For extra measures, I'll delete the cache entry as well if scopes are mismatched, in case user doesn't go through the resign process (also added a unit test to confirm this behavior)

So does this mean that clients don't need to check accessToken if it has all the scopes that we require? Atm we do that since in the past we had an issue with this

hl662:
> > > how do we handle when new scopes are added to a client and they need to resign in?
> > 
> > 
> > The TokenStore compares the new scopes with the scopes in the cached tokenResponse. If different, the load() returns undefined, prompting the user to sign in again. For extra measures, I'll delete the cache entry as well if scopes are mismatched, in case user doesn't go through the resign process (also added a unit test to confirm this behavior)
> 
> So does this mean that clients don't need to check accessToken if it has all the scopes that we require? Atm we do that since in the past we had an issue with this
> 
> cc @DenisMicha

I'm not too sure with that, what the check does is compare if the 2 scopes are different, not necessarily if you have all the scopes your application needs for its purposes). Lastly, this comparison check only exists for the node-cli authentication clients, as far as I can tell.

Issue #166, opened 2023-05-04
Move to ESM


ben-polinsky:
What were previous issues? 

Relates to: https://github.com/iTwin/auth-clients/issues/162

paulius-valiunas:
the main issue I see is you can't use `got@12` from CJS. There might be some more packages too.

PR #203, opened 2023-08-23
Move node-cli to be pure ESM, use Conf for persistent storage

hl662:
- Modify Node-cli's `tsconfig.json` and `package.json` to emit ESM on build 
- Replace `node-persist` with [Conf](https://github.com/sindresorhus/conf), very similar to `electron-store` used by `@itwin/electron-authorization`
- Addresses #166

### DIFF
--- a/packages/node-cli/package.json
+++ b/packages/node-cli/package.json
@@ -20,7 +20,7 @@
     "rebuild": "npm run clean && npm run build"
   },
   "exports": "./lib/esm/index.js",
-    "typings": "./lib/esm/index.d.ts",
+  "typings": "./lib/esm/index.d.ts",
   "author": {
     "name": "Bentley Systems, Inc.",
     "url": "http://www.bentley.com"


packages/node-cli/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
   "compilerOptions": {
-    "target": "es2021",
-    "module": "commonjs",
+    "target": "ES2022",
+    "module": "Node16",
+    "moduleResolution": "Node16",
```
aruniverse, line 6: should this be `NodeNext`

packages/node-cli/package.json
```diff
@@ -1,22 +1,27 @@
 {
   "name": "@itwin/node-cli-authorization",
   "version": "1.0.1",
-  "main": "lib/cjs/index.js",
-  "types": "lib/cjs/index.d.ts",
+  "type": "module",
   "description": "Node.js command-line authorization client for iTwin platform",
+  "engines": {
+    "node": ">=16"
```
aruniverse, line 7: node16 reaches eol in 3 weeks, wonder if we just make the jump to node18

aruniverse: if we do that, we should drop core 3.x deps

packages/node-cli/package.json
```diff
@@ -1,22 +1,27 @@
...
     "docs": "RUSHSTACK_FILE_ERROR_BASE_FOLDER='../..' betools docs --includes=../../generated-docs/extract --json=../../generated-docs/auth-clients/node-cli-authorization/file.json --tsIndexFile=./index.ts --onlyJson",
     "lint": "eslint -f visualstudio \"./src/**/*.ts\" 1>&2",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha \"./lib/cjs/test/**/*.test.js\"",
+    "test": "mocha \"./lib/esm/test/**/*.test.js\"",
     "cover": "nyc npm test",
     "pack": "npm pack",
     "rebuild": "npm run clean && npm run build"
   },
+  "main": "./lib/esm/index.js",
```
aruniverse, line 22: doesnt main point to the cjs ver which this is not so we dont want this?
should this instead point be `module` not `main`?

packages/node-cli/src/Client.ts
```diff
@@ -11,15 +11,15 @@
...
+import open from "open";
 import { assert, BeEvent, BentleyError, Logger } from "@itwin/core-bentley";
 import {
   AuthorizationError, AuthorizationNotifier, AuthorizationRequest, AuthorizationRequestHandler, AuthorizationResponse,
   AuthorizationServiceConfiguration, BaseTokenRequestHandler, BasicQueryStringUtils, GRANT_TYPE_AUTHORIZATION_CODE, GRANT_TYPE_REFRESH_TOKEN,
   TokenRequest,
 } from "@openid/appauth";
-import { NodeCrypto, NodeRequestor } from "@openid/appauth/built/node_support";
-import { TokenStore } from "./TokenStore";
+import { NodeCrypto, NodeRequestor } from "@openid/appauth/built/node_support/index.js";
```
aruniverse, line 21: do we really need to modify imports of our deps to point to the js file?

hl662: unfortunately, yes...
Relative imports need to explicitly have the file extensions `.js` attached to them when we set module and moduleresolution to `Node16` or `NodeNext`. That was a rule seemingly defined by the TS team - what the TS linter does is try to resolve the import to a specific file. If it can't find a specific file, it adds a .js extension (this is what the `node` resolver does).

Specifying `--experimental-specifier-resolution=node` would have worked, bypassing the TS linter rules, but it's experimental, and starting from Node 19 this flag was dropped :(

See the comments on (this StackOverflow post)[https://stackoverflow.com/questions/62619058/appending-js-extension-on-relative-import-statements-during-typescript-compilat] on an explanation on what goes on underneath the compiler, and see (this github discussion)[https://github.com/microsoft/TypeScript/issues/42151] on the TS devs being tired of explaining their reasoning 😢

packages/node-cli/package.json
```diff
@@ -33,15 +32,15 @@
...
   "dependencies": {
-    "@itwin/core-common": "^3.3.0 || ^4.0.0",
+    "@itwin/core-common": "^4.0.0",
     "@openid/appauth": "^1.3.1",
     "conf": "^11.0.2",
     "open": "^8.3.0",
     "username": "^5.1.0"
   },
   "devDependencies": {
     "@itwin/build-tools": "^4.0.0-dev.93",
```
aruniverse, line 42: this should be ^4.0.0 as well

aruniverse: build tools that is

hl662: done

packages/node-cli/package.json
hl662: yes, removed. We don't need `module`, the `exports` field is the official way to do this, per the Node documentation

packages/node-cli/package.json
hl662: bumped all core deps to ^4.0.0

packages/node-cli/package.json
```diff
@@ -1,22 +1,26 @@
...
     "lint": "eslint -f visualstudio \"./src/**/*.ts\" 1>&2",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha \"./lib/cjs/test/**/*.test.js\"",
+    "test": "mocha \"./lib/esm/test/**/*.test.js\"",
     "cover": "nyc npm test",
     "pack": "npm pack",
     "rebuild": "npm run clean && npm run build"
   },
+  "exports": "./lib/esm/index.js",
+    "typings": "./lib/esm/index.d.ts",
```
aruniverse, line 23: indentation

packages/node-cli/tsconfig.json
hl662: after you mentioned node16 is hitting eol, I agree, just changed all of these to nodenext

aruniverse: node.js v16 eol is separate from ts node16 module resolutions; for the latter you can read up on it at https://www.typescriptlang.org/tsconfig#moduleResolution

packages/node-cli/src/Client.ts
hl662: Reading all the comments about this, it's starting to make sense... including the explicit file extension improves module resolution performance (the loader doesn't need to go through every permutation of `.js` or `.d.js` or `.ejs`), and it reduces ambiguity. The thing is, not having the file extension was how the majority of devs have done it in the past, so while including it is the recommended, it's doesn't help with uniformity

PR #215, opened 2023-10-27
Reusable browser automation

MichaelBelousov:
- extract playwright automation of login page out of `TestBrowserAuthorizationClient` and into independent `SignInAutomation` module in `@itwin/oidc-signin-tool`.
- add `loginElectronMainAuthClient` function module to in `@itwin/oidc-signin-tool`, with a high level usage of `SignInAutomation` for electron apps, including those that run iTwin.js backends in an electron utility process
- make playwright a peer dependency of `@itwin/oidc-signin-tool` to prevent double requiring it... this is a breaking change I suppose so I will consider how to remove it
- (attempting to remove this) make electron an optional (for types) dependency of `@itwin/oidc-signin-tool`


ben-polinsky:
Adding a peer dep is definitely a breaking change, but I don't see a problem with that. Is it really optional though? 

> with a high level usage of SignInAutomation for electron apps, including those that run iTwin.js backends in an electron utility process

I just want to make sure these changes are intended for testing and not for other automation.

MichaelBelousov:
> Adding a peer dep is definitely a breaking change, but I don't see a problem with that. Is it really optional though?
> 
> > with a high level usage of SignInAutomation for electron apps, including those that run iTwin.js backends in an electron utility process
> 
> I just want to make sure these changes are intended for testing and not for other automation.

yes they are solely for testing. I will think about how to remove the peer dep... it is necessary for the case in which a consumer of the library brings their own playwright. You can't load two instances of playwright into the same node.js process, so then this package couldn't use its own dependency and the package manager has to place things correctly.

I suppose I might be able to do it by dynamically `require`ing playwright at usage sites throughout this package, that way if you use an API that requires passing your own playwright `Page`, then this package will not attempt to load its own playwright. But I will need to test consuming this from a consumer both with and without playwright

ben-polinsky:
> yes they are solely for testing

Cool. 

Yeah the double playwright issue has been annoying with this tool. I think that adding it as a non-optional peer dep could be okay. I don't mind a breaking change if it clears this up once and for all.

MichaelBelousov:
@ben-polinsky bump? we just got this again in CI.

Also, I was thinking about the peer dependency... it _might_ be possible to install playwright if it can't be found, as a postinstall script, but I'm really not sure.

aruniverse (approved):

PR #217, opened 2023-10-27
Update oidc-signin-ts to address GHSA-xwcq-pm8m-c4vf

nick4598:
https://github.com/advisories/GHSA-xwcq-pm8m-c4vf


nick4598:
Is it a problem that my lockfile is version 6.0 instead of 5.4?

aruniverse (approved):

Issue #224, opened 2023-12-09
`@itwin/oidc-signin-tool` password is not always filled in


johnnyd710:
Sometimes (not all the time) in our MacOS Github Action, the password is not filled out in the sign-in form using `@itwin/oidc-signin-tool` with our Electron application. I was able to catch a video of it after a few attempts. In the third second of the video you will see "Please fill out this field." and the password is not filled in, I attached a screenshot of the moment.

![image](https://github.com/iTwin/auth-clients/assets/19596966/10ce904c-1250-414c-bd7f-cf7a7e0fc068)

https://github.com/iTwin/auth-clients/assets/19596966/a9dbd2c2-bbd0-44a4-853a-f349541cf578

Reach out to me if you have any questions.

johnnyd710:
I patched the `oidc-signin-tools` package to take a screenshot and dump the html to a file right before we enter the password. Problem is, I can't reproduce anymore after 4 attempts.

Its possible there is a race condition, and the time it takes to take the screenshot and dump the html to a file fixes the problem.

MichaelBelousov:
4 attempts doesn't seem like enough based on what I know about our CI...
but yes the HTML probably hurts the odds, perhaps significantly... can you make the CI try it on loop more times?

johnnyd710:
I *think* replacing the `.type` calls with `.fill` helped, my CI keeps failing for other reasons so its hard to be sure, but I haven't seen the error in a few runs with repeatEach set to 5

PR #225, opened 2023-12-11
Replace playwright `type` calls with `fill`

johnnyd710:
Using `locator.fill` is [the recommended way](https://playwright.dev/docs/input#text-input) to fill out input, and [`Page.type` is now deprecated](https://playwright.dev/docs/api/class-page#page-type).

I made these changes using `pnpm patch` for my Electron application, and I *think* it will close https://github.com/iTwin/auth-clients/issues/224 but I can't be sure because the issue is flaky and doesn't happen frequently.

### DIFF
--- a/change/@itwin-oidc-signin-tool-ca6fd177-8f76-4fe4-afb8-ea94ece5d83c.json
+++ b/change/@itwin-oidc-signin-tool-ca6fd177-8f76-4fe4-afb8-ea94ece5d83c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Refactor login form handling in SignInAutomation.ts to use .fill",
+  "packageName": "@itwin/oidc-signin-tool",
+  "email": "19596966+johnnyd710@users.noreply.github.com",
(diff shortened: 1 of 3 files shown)


aruniverse (approved):

ben-polinsky:
Looks good, but integration tests are now failing. Could you look into why?

johnnyd710:
> Looks good, but integration tests are now failing. Could you look into why?

The `basic.test.t > failure with invalid Bentley federated user` integration test was entering the invalid@bentley.com username twice (another reason not to use `type`) and the test was checking for `Enter a valid email address, phone number, or Skype name.` which is the incorrect thing to test for.
image

It should be checking for `This username may be incorrect. Make sure you typed it correctly.` for the case of an invalid username but correctly formatted email, like it does now that `fill` replaces the input text instead of adding to it. 

image

johnnyd710:
Now its failing on mac because `ErrorResponse: Authorization code is invalid or expired.` ?

ben-polinsky:
Thanks @johnnyd710

Issue #226, opened 2023-12-12
`ElectronMainAuthorization.signIn` throws uncatchable error when `tokenUrl` server errors


johnnyd710:
Execution flow for our Electron App:
- The frontend listens to `ElectronRendererAuthorization.onAccessTokenChanged` and calls `await ElectronRendererAuthorization.signIn()`.
- The backend `ElectronMainAuthorization` kicks off a sign in request.
- The frontend `signIn()` promise resolves, and a "signing in..." page is shown
- User enters username and password, logs in successfully to get an Authorization Code.
- While swapping the Authorization Code for a token, the `tokenUrl` server (called inside `ElectronMainAuthorization.swapAuthorizationCodeForTokens`) throws a `500 Internal Server Error` (happens rarely, but enough to be a bother in our CI).
- Our backend throws the error and but frontend is left hanging on the "signing in..." page because `onAccessTokenChanged` is never called.

Ideally, I'd like the `signIn()` to resolve or reject once the entire sign in process is complete, not just after starting the process, that way our frontend can properly react to events or errors. If not that is not possible, at the very least there should at least be a way for our backend to catch the `signIn()` errors initiated through IPC by the frontend, or an event we can hook into, so we can tell the frontend ourselves.

Happy to look into this.

ben-polinsky:
> Happy to look into this.

Sure, and thanks for the report.

PR #227, opened 2023-12-13
Resolve `signIn` promise when sign in has completed

johnnyd710:
tldr: Resolve the `ElectronRendererAuthorization.signIn()` promise when sign-in has fully completed, and reject on errors. Closes https://github.com/iTwin/auth-clients/issues/226.

### Changes
Previously, some sign-in errors (e.g., an internal server error on the token issuer server) were uncaught by `ElectronRendererMain`, and there wasn't a way for consumers of this package to catch it. The frontend who initiated the sign-in was left waiting for `onAccessTokenChanged` forever. Now, we resolve the promise when sign in has actually fully completed or reject the promise if there was an error during sign in, so that the frontend can show an error message.

Technically the API is the same, the types are the same, but the usage could vary slightly.

### Before
```jsx
  const onSignInClick = async (): Promise<void> => {
    setSigningIn(true);
    if (IModelApp.authorizationClient instanceof ElectronRendererAuthorization) {
      await IModelApp.authorizationClient.signIn(); // resolves as soon as sign-in starts, with no way to catch errors
    }
  };

  useEffect(() => {
    (IModelApp.authorizationClient as ElectronRendererAuthorization)?.onAccessTokenChanged.addListener(
      (token: AccessToken | undefined) => {
        setAccessToken(token);
        setSigningIn(false);
      }
    );
  });

```

### After
```jsx
  const onSignInClick = async () => {
    setSigningIn(true);
    if (IModelApp.authorizationClient instanceof ElectronRendererAuthorization) {
      try {
        await IModelApp.authorizationClient.signIn(); // resolves after sign in is complete
      } catch (err) {
        LoggingClient.logException(err); // catch errors and do something
        setErrorMessage("Error while signing in, please try again.");
        setSigningIn(false);
      }
    }
  };

  useEffect(() => {
    (IModelApp.authorizationClient as ElectronRendererAuthorization)?.onAccessTokenChanged.addListener(
      (token: AccessToken | undefined) => {
        setAccessToken(token);
        setSigningIn(false);
      }
    );
  });
```

### DIFF
--- a/change/@itwin-electron-authorization-cb30a53e-3752-4c76-829e-b7017135452b.json
+++ b/change/@itwin-electron-authorization-cb30a53e-3752-4c76-829e-b7017135452b.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Resolve ElectronRendererAuthorization signIn when sign in has completed and reject if an error occurs.",
+  "packageName": "@itwin/electron-authorization",
+  "email": "19596966+johnnyd710@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


aruniverse (approved):
makes sense

ben-polinsky:
LGTM, thanks.

ben-polinsky (approved):

PR #228, opened 2023-12-14
Publish static loopback pages

johnnyd710:
Static loopback pages needed to be published to npm, they were ignored by the `.npmignore` file, resulting in a runtime `ENOENT: no such file or directory`


ben-polinsky:
Good catch - thanks.

aruniverse (approved):
No idea how this worked without this

ben-polinsky (approved):

PR #234, opened 2024-03-14
Add org to introspection response

andrewdinunzio:


aruniverse (commented):
missing beachball change

andrewdinunzio:
@aruniverse Sorry could you explain? First time working with this

andrewdinunzio:
I chose "backwards-compatible change" and it showed up as patch in the change file, hmm

aruniverse (approved):

PR #235, opened 2024-03-15
Add EU cookie consent banner handling to OIDC sign in

jsnaras:
In EU the cookie consent banner is obscuring the "Accept" button and must be dismissed before proceeding.t

image


aruniverse (approved):
god bless the EU and their policies

Issue #241, opened 2024-05-06
`process` is not available in the browser


veekeys:
<img width="564" alt="image" src="https://github.com/iTwin/auth-clients/assets/81580355/6154e85b-a566-46fa-a526-bd72e3493ddb">

Getting this with the newest `@itwin/browser-authorization` version. Probably, should not rely on environment variables for the resolution of authority...

ben-polinsky:
Thanks. 

This is due to our earlier reliance on webpack/cra magic and has not been caught as the `getImsAuthority` method is only called as a fallback.

ben-polinsky:
@veekeys Could you test out the fix in `v1.1.1`?

ben-polinsky:
Closing, please let us know if the issue is not resolved.

veekeys:
@ben-polinsky Sorry, checked now, seems that error is gone.
Thanks!

ben-polinsky:
np! Thanks for testing.

PR #242, opened 2024-05-07
[Browser] Deprecate getImsAuthority and usage of process.env 

ben-polinsky:
#241


packages/browser/src/utils.ts
```diff
@@ -11,16 +11,22 @@
...
-  // "dev-ims.bentley.com" doesn't exist, so convert prefix to "qa-"
-  if (prefix === "dev-")
-    prefix = "qa-";
+    return `https://${prefix}ims.bentley.com`;
+  } catch (_) {
+    // swallow error
+  }
 
-  return `https://${prefix}ims.bentley.com`;
+  return `https://ims.bentley.com`;
```
aruniverse, line 31:
```suggestion
  return "https://ims.bentley.com";
```

PR #243, opened 2024-05-07
upgrade docs build tools, typescript, and eslint

ben-polinsky:
Upgrade @itwin/build-tools to pre-release to take advantage of new typedoc version. TS bumped to 5.3.3. ESlint upgraded to 8.x. Upgrade @itwin/eslint-plugin to 4.0.x stable.

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -35,9 +35,9 @@
     "oidc-client-ts": "^2.4.0"
   },
   "devDependencies": {
-    "@itwin/build-tools": "^4.0.0-dev.93",
+    "@itwin/build-tools": "^4.6.0-dev.27",
     "@itwin/core-bentley": "^3.7.0",
-    "@itwin/eslint-plugin": "^3.7.0",
+    "@itwin/eslint-plugin": "^4.0.1",
     "@playwright/test": "~1.41.0",
     "@types/chai": "^4.2.22",
     "@types/mocha": "^8.2.3",
@@ -46,26 +46,16 @@
     "buffer": "~6.0.3",
     "chai": "^4.2.22",
     "dotenv": "~16.0.3",
-    "eslint": "^7.32.0",
+    "eslint": "^8.56.0",
     "mocha": "^8.2.3",
     "nyc": "^15.1.0",
     "parcel": "~2.12.0",
     "process": "~0.11.10",
     "rimraf": "^3.0.2",
     "sinon": "^15.0.1",
-    "typescript": "~5.0.2"
+    "typescript": "~5.3.3"
   },
   "peerDependencies": {
     "@itwin/core-bentley": "^3.3.0 || ^4.0.0"
-  },
-  "eslintConfig": {
-    "plugins": [
-      "@itwin"
-    ],
-    "extends": "plugin:@itwin/itwinjs-recommended",
-    "rules": {
-      "no-duplicate-imports": "off",
-      "@typescript-eslint/consistent-type-imports": "error"
-    }
   }
 }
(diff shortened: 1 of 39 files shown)


ben-polinsky:
We need a fix from our eslint library for this to lint correctly: 

[PR](https://github.com/iTwin/eslint-plugin/pull/67) - [Commit](https://github.com/iTwin/eslint-plugin/pull/67/commits/485f27dbe9cc1852367e5b68cab0835377aa0d5c)

packages/browser/package.json
```diff
@@ -35,9 +35,9 @@
     "oidc-client-ts": "^2.4.0"
   },
   "devDependencies": {
-    "@itwin/build-tools": "^4.0.0-dev.93",
+    "@itwin/build-tools": "^4.6.0-dev.27",
```
aruniverse, line 38: any reason to use pre-release?

packages/browser/src/types.ts
```diff
@@ -44,14 +44,7 @@ export interface BrowserAuthorizationClientConfiguration
   /** A space-delimited collection of individual access claims specified by the authority. The user must consent to all specified scopes in order to grant authorization */
   readonly scope: string;
   /** The mechanism (or authentication flow) used to acquire auth information from the user through the authority */
-  readonly responseType?:
-  | "code"
```
aruniverse, line 48: what was the reason for this?
wouldnt this make it harder for consumers to know what are actual valid response types?
not that the string union at the end helped too much

packages/oidc-signin-tool/src/SignInAutomation.ts
```diff
@@ -26,7 +26,8 @@ interface AutomatedContextBase<T> {
   /** A function that takes the waitForCallback result data (e.g. a callback url)
    * and finalizes the sign in process
    */
-  resultFromCallback?: (t: T) => any | Promise<any>;
+  // eslint-disable-line @typescript-eslint/no-redundant-type-constituents
```
aruniverse, line 29: do you need both these disables?
should this be disable next line?

aruniverse (approved):

packages/browser/package.json
ben-polinsky: Yes, we haven't yet published a version with new typedoc

packages/browser/src/types.ts
ben-polinsky: `@typescript-eslint/no-redundant-type-constituents : "code" is overridden by string in this union type.`
https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHIBGcUAXMgEQkBelyAPsgM5hSgDmuAvrrguhCtkMChiwBeHPkIlyyHkA - no autocomplete.

We could provide suggested valid values in the comment, or if we think we know all possible values, we could remove the `string`.

packages/oidc-signin-tool/src/SignInAutomation.ts
ben-polinsky: At one point I thought I did, but re-tested and not needed. Updated.

packages/browser/src/types.ts
ben-polinsky: https://openid.net/specs/openid-connect-core-1_0.html#Authentication

aruniverse: im not an auth expert so dont know all the other possible response types.
if this is the exhaustive list, then id rather remove the string

@evelynpreslar-bentley ?

aruniverse: ah thx for sharing ben, looks like thats the list, and we can remove string

Issue #246, opened 2024-05-09
Upgrade to v3 of oidc-client-ts


ben-polinsky:
https://github.com/authts/oidc-client-ts/blob/main/docs/migration.md

ben-polinsky:
This is a breaking change for oidc client because users on < node 18 will have to add a polyfil for crypto.

PR #251, opened 2024-06-18
node 20; oidc-ts-client 3.x; remove itwin- 3.x packages

ben-polinsky:
- Update ci and node types to 20.x. Not only is 20.x LTS, but it is required for oidc-ts-client@3.x which relies on the browser's crypto. Node>18.x automatically adds `crypto` to `globalThis`. 
- update oidc-ts-client to 3.x
- remove support for itwin 3.x packages

closes: #246

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -31,17 +31,17 @@
     "directory": "packages/browser"
   },
   "dependencies": {
-    "@itwin/core-common": "^3.7.0 || ^4.0.0",
-    "oidc-client-ts": "^2.4.0"
+    "@itwin/core-common": "^4.0.0",
+    "oidc-client-ts": "^3.0.1"
   },
   "devDependencies": {
-    "@itwin/build-tools": "^4.6.0-dev.27",
-    "@itwin/core-bentley": "^3.7.0",
-    "@itwin/eslint-plugin": "^4.0.2",
+    "@itwin/build-tools": "^4.7.1",
+    "@itwin/core-bentley": "^4.7.1",
+    "@itwin/eslint-plugin": "^4.1.0",
     "@playwright/test": "~1.41.0",
     "@types/chai": "^4.2.22",
     "@types/mocha": "^8.2.3",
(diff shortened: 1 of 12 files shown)


packages/browser/package.json
```diff
@@ -56,6 +56,6 @@
     "typescript": "~5.3.3"
   },
   "peerDependencies": {
-    "@itwin/core-bentley": "^3.3.0 || ^4.0.0"
+    "@itwin/core-bentley": "^4.0.0"
```
aruniverse, line 59: are we going to make this a major bump?

packages/browser/package.json
```diff
@@ -31,17 +31,17 @@
...
-    "oidc-client-ts": "^2.4.0"
+    "@itwin/core-common": "^4.0.0",
+    "oidc-client-ts": "^3.0.1"
   },
   "devDependencies": {
-    "@itwin/build-tools": "^4.6.0-dev.27",
-    "@itwin/core-bentley": "^3.7.0",
-    "@itwin/eslint-plugin": "^4.0.2",
+    "@itwin/build-tools": "^4.7.1",
+    "@itwin/core-bentley": "^4.7.1",
```
aruniverse, line 39: does core-bentley need to be in lockstep with the other core pkgs?

packages/browser/package.json
ben-polinsky: Yep that is the idea

packages/browser/package.json
ben-polinsky: Are you suggesting they should both be 4.7.0? Thought for dev deps our policy was to bump to latest compatible with deps.

ben-polinsky:
Don't know that it's worth moving to pure esm until core is there...

ben-polinsky:
> Don't know that it's worth moving to pure esm until core is there...

After discussion with @aruniverse, we'll pick this back up closer to core@5 release. We'll go full ESM at that point as well.

PR #253, opened 2024-07-02
Add missing support for subpaths

hl662:
When using `export` and subpaths, to fully support CJS and ESM, we need each subpath to contain `import` and `require`.

Previously, `./Renderer` didn't support CJS, because it didn't have a `require` field. This would cause problems in packages, like `itwinjs/core` `core-full-stack-tests`, that consumes both Main and Renderer, but outputs CJS files. At build time, this wouldn't be caught, but during runtime, it would throw a `PACKAGE_PATH_NOT_SUPPORTED`, because when it tried to use the Renderer subpath, it couldn't find a valid CJS path.

Vice versa, with `./Main` in ESM outputted files.

This issue is fixed by adding both `import` and `require` and their valid values to each subpath.


aruniverse (approved):

ben-polinsky:
Excellent. I ran into this setting up my docs cospace the other day. Thanks!

ben-polinsky (approved):

PR #259, opened 2024-10-18
Add support for Electron 33

GytisCepk:
[Electron 33 breaking changes](https://github.com/electron/electron/blob/main/docs/breaking-changes.md#planned-breaking-api-changes-330)

### DIFF
--- a/change/@itwin-electron-authorization-06682ad1-2406-4ed3-ad13-cccf7c0b8128.json
+++ b/change/@itwin-electron-authorization-06682ad1-2406-4ed3-ad13-cccf7c0b8128.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add support for Electron 33",
+  "packageName": "@itwin/electron-authorization",
+  "email": "GytisCepk@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "Authorization clients for the iTwin platform",
   "scripts": {
     "build": "lage build",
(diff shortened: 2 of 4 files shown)


ben-polinsky (approved):

package.json
```diff
@@ -5,7 +5,7 @@
   "description": "Authorization clients for the iTwin platform",
   "scripts": {
     "build": "lage build",
-    "clean": "lage clean",
+    "clean": "lage clean --no-cache",
```
anmolshres98, line 8: is this needed?

change/@itwin-electron-authorization-06682ad1-2406-4ed3-ad13-cccf7c0b8128.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
anmolshres98, line 2: should this be a minor type?

GytisCepk: Electron package is still in 0.x version. This means for fixes and new features we should do patch release, minor bumps are for breaking changes.

package.json
GytisCepk: Cacheable `clean` script is a bit comical. If you haven't changed source files since last `clean` script run, cli output will be replayed, but nothing will get deleted.

I wanted to put this inside lage config file, but couldn't find a way to make it work.

anmolshres98: sounds like a huge bug, smh

anmolshres98 (approved):

aruniverse (approved):

Issue #262, opened 2024-10-25
bug: electron app e2e test hangs while using `oidc-signin-tool`


johnnyd710:
The test hangs at this line: https://github.com/iTwin/auth-clients/blob/main/packages/oidc-signin-tool/src/SignInAutomation.ts#L169. Commenting out *all* 'networkidle' lines in a patch fix solves the issue. 

Playwright's [waitForLoadState('networkidle') is discouraged](https://playwright.dev/docs/api/class-page#page-wait-for-load-state). I think the package should be refactored to avoid using it.

johnnyd710:
@MichaelSwigerAtBentley can we close this now? seems to be working.

MichaelSwigerAtBentley:
Yes. There are still some 'networkidle' calls in the integration tests that I would like to remove, but I will create a new issue to track that.

PR #265, opened 2024-10-28
Remove Discouraged 'networkidle' Calls

MichaelSwigerAtBentley:
Removes 4 waitForLoadState('networkidle') calls in SignInAutomation.ts as outlined in issue #262


MichaelSwigerAtBentley:
I was unable to get integration tests running locally (token requests failed with invalid client Id), even with the same configs and @ben-polinsky's help. Opening a PR for now to try integration tests through the pipeline.

PR #266, opened 2024-10-28
Remove Discouraged 'networkidle' Calls

MichaelSwigerAtBentley:
Removes 4 waitForLoadState('networkidle') calls in SignInAutomation.ts as outlined in issue https://github.com/iTwin/auth-clients/issues/262

I was unable to get integration tests running locally (token requests failed with invalid client Id), even with the same configs and @ben-polinsky's help. Opening a PR for now to try integration tests through the pipeline.

### DIFF
--- a/packages/oidc-signin-tool/src/SignInAutomation.ts
+++ b/packages/oidc-signin-tool/src/SignInAutomation.ts
@@ -111,7 +111,6 @@ export async function automatedSignOut<T>(
 }
 
 async function handleErrorPage<T>({ page }: AutomatedContextBase<T>): Promise<void> {
-  await page.waitForLoadState("networkidle");
   const pageTitle = await page.title();
   let errMsgText;
 
@@ -166,7 +165,6 @@ async function handlePingLoginPage<T>(context: AutomatedSignInContext<T>): Promi
   allow = page.locator(testSelectors.pingAllowSubmit);
   await allow.click();
 
-  await page.waitForLoadState("networkidle");
   const error = page.getByText(
     "We didn't recognize the email address or password you entered. Please try again.",
   );
@@ -187,7 +185,6 @@ async function handlePingLoginPage<T>(context: AutomatedSignInContext<T>): Promi
 async function handleFederatedSignin<T>(context: AutomatedSignInContext<T>): Promise<void> {
   const { page } = context;
 
-  await page.waitForLoadState("networkidle");
   if (-1 === page.url().indexOf("microsoftonline"))
     return;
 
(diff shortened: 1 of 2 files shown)


packages/oidc-signin-tool/src/SignInAutomation.ts
```diff
@@ -187,10 +185,14 @@ async function handlePingLoginPage<T>(context: AutomatedSignInContext<T>): Promi
 async function handleFederatedSignin<T>(context: AutomatedSignInContext<T>): Promise<void> {
   const { page } = context;
 
-  await page.waitForLoadState("networkidle");
   if (-1 === page.url().indexOf("microsoftonline"))
     return;
 
+  // Wait for either msUserNameField or fedPassword to be visible
+  const msUserNameFieldPromise = page.waitForSelector(testSelectors.msUserNameField, { state: "visible" });
+  const fedPasswordPromise = page.waitForSelector(testSelectors.fedPassword, { state: "visible" });
```
ben-polinsky, line 193: What's the point of optionally waiting for this field if it isn't used?

ben-polinsky:
There are also several `waitForLoadState('networkidle')` calls in the electron package's integration tests. If this fix is needed asap, we could do that in a separate PR.

packages/oidc-signin-tool/src/SignInAutomation.ts
MichaelSwigerAtBentley: Because its possible that the userNameField will be skipped and go straight to the passwordField, so just waiting for the userNameField to be visible wont be enough, correct?

MichaelSwigerAtBentley: But we need to wait for something here (previously networkidle) before we attempt to check if the userNameField is there

MichaelSwigerAtBentley:
> There are also several `waitForLoadState('networkidle')` calls in the electron package's integration tests. If this fix is needed asap, we could do that in a separate PR.

Yes, 

> There are also several `waitForLoadState('networkidle')` calls in the electron package's integration tests. If this fix is needed asap, we could do that in a separate PR.

I agree that that should be a separate PR, once we verify that this is a valid fix for the e2e tests

packages/oidc-signin-tool/src/SignInAutomation.ts
ben-polinsky: Ohh right, I forgot about that case. All good.

ben-polinsky (approved):

aruniverse (approved):

PR #272, opened 2024-11-15
Update to latest version of username

wgoehrig:
Older versions of the `username` package have an (indirect, via `execa`) dependency on `cross-spawn` which has a newly-reported ReDoS CVE (https://nvd.nist.gov/vuln/detail/CVE-2024-21538).

Unfortunately, the newest version of `username` is ESM-only.  However, I believe with a few minor tsconfig changes, we should be able to safely dynamic-import `username` in a backwards-compatible way, since `import()` is supported in all our currently-[supported platforms](https://www.itwinjs.org/learning/supportedplatforms/).

### DIFF
--- a/packages/electron/src/main/TokenStore.ts
+++ b/packages/electron/src/main/TokenStore.ts
@@ -4,7 +4,6 @@
 *--------------------------------------------------------------------------------------------*/
 // Code based on the blog article @ https://authguidance.com
 
-import * as OperatingSystemUserName from "username";
 import { safeStorage } from "electron";
 // eslint-disable-next-line @typescript-eslint/naming-convention
 const Store = require("electron-store"); // eslint-disable-line @typescript-eslint/no-var-requires
@@ -37,13 +36,7 @@ export class RefreshTokenStore {
 
   private async getUserName(): Promise<string | undefined> {
     if (!this._userName) {
-      try {
-        this._userName = await OperatingSystemUserName();
-      } catch {
-        // errors occur in testing when using asynchronous call
-        // https://github.com/iTwin/auth-clients/issues/163
-        this._userName = OperatingSystemUserName.sync();
-      }
+      this._userName = await (await import("username")).username();
     }
 
     return this._userName;
--- a/packages/electron/tsconfig.json
+++ b/packages/electron/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
   "compilerOptions": {
-    "target": "es2022",
+    "target": "es2021",
   },
   "include": [
     "./src/**/*.ts"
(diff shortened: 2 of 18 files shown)


packages/electron/src/main/TokenStore.ts
```diff
@@ -4,7 +4,6 @@
 *--------------------------------------------------------------------------------------------*/
 // Code based on the blog article @ https://authguidance.com
 
-import * as OperatingSystemUserName from "username";
 import { safeStorage } from "electron";
 // eslint-disable-next-line @typescript-eslint/naming-convention
 const Store = require("electron-store"); // eslint-disable-line @typescript-eslint/no-var-requires
```
aruniverse, line 9: can we move away from the require here?

packages/electron/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
   "compilerOptions": {
-    "target": "es2021",
+    "target": "es2022",
```
aruniverse, line 4: do we need to change the ver bump because of this?

packages/electron/src/main/TokenStore.ts
wgoehrig: not sure what you mean... if you mean do we rely on a side-effect from this require, I think the answer is no...

packages/electron/tsconfig.json
wgoehrig: Actually I should be able to revert these and stick at es2021

ben-polinsky (approved):

packages/electron/src/main/TokenStore.ts
aruniverse: ah missed the side effect

aruniverse (approved):

ben-polinsky:
@wgoehrig were you able to merge this without the docs build passing or did you force?

wgoehrig:
> @wgoehrig were you able to merge this without the docs build passing or did you force?

@ben-polinsky I just had this set to auto-merge - I didn't bypass anything

wgoehrig:
Looks like the docs pipeline was missing from the branch policy - just added that so this doesn't happen again

aruniverse:
> That being said, I have no idea what I did in this PR to break the docs build... ☹️
> 
> @ben-polinsky any ideas?

Looks like docs are actually incorrect in core? and its not an auth client repo docs issue?

ben-polinsky:
> Looks like the docs pipeline was missing from the branch policy - just added that so this doesn't happen again

Thanks! 

As for ideas... maybe some sort of typedoc output version mismatch - trying to track it down now

ben-polinsky:
> Looks like docs are actually incorrect in core? and its not an auth client repo docs issue?
Looks okay to me... the failures I see in core are audit issues. The error the link is referencing doesn't really seem to be an error.

PR #281, opened 2025-01-15
Update "Releases" page when a new tag is pushed to repo

### LINKED ISSUE #229: Releases are not being added to GitHub
We've released many times since this:

image

anmolshres98:
Closes #229 

This PR makes it so that every time a package is published to npm and a tag is created then pushed to this repo, the newly created script
- discerns the packageName and version from the tag
- determines what the package directory is and the CHANGELOG md for the package would be
- extracts the relevant release notes for the current tagged release 
- determines and creates zipped artifacts for the specific package
- uses [gh cli](https://cli.github.com/manual/gh_release_create) to create a new release using the triggered tag containing above parsed release notes and artifacts

Testing:
- one of the missing release created using the new workflow: https://github.com/iTwin/auth-clients/releases/tag/%40itwin%2Fservice-authorization_v1.2.3
- a test release auto-triggered and created by the pipeline when a test tag is pushed to this repo: https://github.com/iTwin/auth-clients/releases/tag/%40itwin%2Ftester_v1.0.1 (I will delete this test release and the test tag once this PR is merged)


.github/workflows/create-github-release.yaml
```diff
@@ -0,0 +1,30 @@
+name: Create GitHub Release
+
+on:
+  push:
+    tags:
+      - "**"
```
ben-polinsky, line 6: Should we target our release tag pattern specifically?

.github/workflows/publish-packages.yaml
```diff
@@ -1,6 +1,4 @@
-# Workflow that will create a GitHub release a when a new release tag is pushed
-
-name: Create Release
```
aruniverse, line 3: do we need to rename the file and the name of the task?
does this preserve action history in the actions tab? and if it does, does it retain the old ones under the old name which can be confusing?

.github/workflows/scripts/create-release.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
aruniverse, line 1: why are we creating new bash scripts?
we've been moving away from this in all our other repos
eg in core if we need to run a script we will write in js and use tools like zx

ben-polinsky (approved):

.github/workflows/create-github-release.yaml
anmolshres98: is there a scenario where we would create a tag but not create a release?

.github/workflows/publish-packages.yaml
anmolshres98: no we don't "need" to rename the file and task but I think the cost of doing so is worth it in our case. Creating releases in the ADO land and github can mean different things so I think it is important to have this disambiguation moving forward. "Publishing to npm" should be clearly semantically separated from "releasing to github". I think this PR works on this disambiguation. I am considering the previous state as "buggy" and this PR "fixing" that. Yes, it will retain previous runs under previous name and all new runs will be under a new workflow. The old version and the new version are separated and both have histories. Like so: 
image

.github/workflows/scripts/create-release.sh
anmolshres98: I did consider using `zx` for this work and I have written big parts of the zx js scripts on core repo. For my decision, here are the considerations I made:
- The reason we want to use zx to write js for bash scripts if a process has complicated enough logic that it would be a better dev experience to write using js
- We can think of using zx and using shell script as a spectrum in terms of usecase. For example, if we want to just print "Hello World" to the shell we would use a shell script directly instead of zx. However, if we need to create a tree and traverse it in the shell (for whatever) reason we would use zx. In this usecase I decided that our application is closer to console logging "Hello World" so closer to the shell script part of the spectrum (because the most complicated task is reading and writing file contents). 
- using zx on pipeline has additional overheads too (installing node, installing the npm package, writing in 2 different languages). 

Since the task does not meet the complexity threshold to use js (imo) and we have additional overheads that are relatively costly for the complexity of the task, it makes the most sense to use native shell scripts

.github/workflows/create-github-release.yaml
ben-polinsky: We don't usually, but we might. In general, tags are certainly not 1:1 to releases. I know we can delete them easily, but it might be worth it for the small amount of code.

anmolshres98: ok that makes sense.

I added the narrow pattern and pushed 2 test tags, one matching the pattern and another not matching:
image

the matching pattern triggered the pipeline: https://github.com/iTwin/auth-clients/actions/runs/12812442000

the non-matching tag name did not trigger the pipeline

ben-polinsky: Great, thank you.

anmolshres98:
The `Create GitHub Release` workflow here is being reported as failed as it was forced to fail on purpose during a test. Check this comment for details: https://github.com/iTwin/auth-clients/pull/281#discussion_r1918790130

.github/workflows/publish-packages.yaml
aruniverse: a question to contemplate, does it make sense to decouple npm pkg publishes with gh releases?

ben-polinsky: I would say that screen shot is confusing. If Create Release is old and never to be used again, then I don't like it sitting there.

Not sure I buy the semantic separation argument since the publish to npm action also creates and pushes the git tag. Can't we name the new workflow push-gh-release or something?

I think the switcheroo is a bit confusing.

.github/workflows/scripts/create-release.sh
MichaelSwigerAtBentley: I think that regardless of complexity, all scripts should follow a consistent workflow. 

I think having multiple script languages makes it harder for an outsider that is unfamiliar with our processes know how to add to and maintain the scripts. I definitely think there is a time where a script can be so simple that js is totally overkill, but I think those should be isolated to inline scripts in the yaml itself.

DanRod1999: Personally, I've written in js using zx, and I don't really like it for some of the reasons Anmol mentioned. I've been thinking about this a fair bit lately and wondering if JS is the right tool for scripting in general (maybe convo for another day time lol)

However imo I choose writing in JS for these reasons:
- easy of use when reading JSON files
  - I see you use JQ and thats fine, we've used it before in other sh scripts, but I think its an easy argument to make that JS most things JSON related
- JS has a pretty good api for GH related stuff, and I've found it very easy to use instead of zx to perform GH related tasks. For example I think you would use [this ](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) instead of the gh cli tool you used. (i've written scripts with both tools, and I think I prefer this)
  - Note: it is annoying having to install node and the package before every run, which is one of my complaints of zx, but I personally believe its better than zx if you are using it for GH related things.
- JS also has pretty good tools for regex built in, so for file manipulation I've had little trouble with that. So for your uses of Sed might have easier workarounds in JS then you think. At least I was surprised when i started doing file manipulation in JS and I realized it was as painful as I expected. So using JS regex with vars might not be that bad either compared to awk and sed
- Arun has put a pretty big emphasis on testing in my recent scripts, and I've found it fairly easy to do in JS and Idk how I would be able to do it in something like sh scripts

All in all I don't have a strong opinion on this script particularly, but how I would've done it is maybe if sed was was necessary use a combo of sh commands directly in the pipeline task, and then pass those vars in as cmd args or env vars to the JS file. The only parts I don't fully understand, or know how I would do in JS are your while loop part and the awk script (mostly cause I don't understand the awk lol). The zipping section also seems a bit difficult to do for this. But if it were me I might have made that a separate step outside of this script. We do this in other ADO pipelines I think because they have steps built for zipping files iirc (workflows might have the same?).

ben-polinsky: Agree a consistent workflow makes sense. I like `zx` but it's a dependency with its own dependencies. Granted, here this is only `node,` but now we have gh actions, the os, and zx we need to ensure play well since they all depend on node. Perhaps that's making a mountain out of a molehill. But here's something else: [shell configuration](https://google.github.io/zx/configuration) - I've run into this before with zsh where I needed to set it. But then we're restricting the script to a particular shell?

I think a year ago, I would have said (and I think did say) let's coalesce around `zx`. Now, with AI assistants easing the pain of bash for those who don't know it... I'm not so sure.

DanRod1999:
Didn't look to close into this, but is creating a release using this action step potentially an option?
https://github.com/actions/create-release

we should also be trying to leverage to tools actions have as much as we can I think. Its what we currently in ado at least.

aruniverse:
> Didn't look to close into this, but is creating a release using this action step potentially an option? [actions/create-release](https://github.com/actions/create-release)
> 
> we should also be trying to leverage to tools actions have as much as we can I think. Its what we currently in ado at least.

I agree we should look at using existing tools instead of rolling our own if possible.
But what Dan linked has been archived and no longer supported, plz dont use it

.github/workflows/scripts/create-release.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Extract the changelog text
+releaseNoteText=$(awk -v version="$packageVersion" '$0 ~ version {flag=1; print; next} /^## /{flag=0} flag' "$changelogMd")
+echo "Release note text was extracted as: $releaseNoteText"
+
+# remove the @itwin/ or @bentley/ scope from the tag name to create the zip file name since they have special characters
+zipFileName=$(echo "$tagName" | sed 's/@itwin\///; s/@bentley\///')
+echo "Zip file name was parsed as: $zipFileName"
+
+# Zip the package directory with just the specific package
```
aruniverse, line 53: wouldnt this just zip up the ts files, dont we want to publish the build output (eg the lib/**.{js,css, etc})

.github/workflows/publish-packages.yaml
aruniverse: ideally this is reverted, however since you already ran this new pipeline, this action shows up and theres no way to remove it. so we now have both workflows showing up, and at this point reverting doesnt help much either besides now keeping history of all the publishes in a single place

.github/workflows/scripts/create-release.sh
aruniverse: ok thanks all for chiming in, im ok with the sh script then. i do hope everyone thinks about the pros & cons before they wright a new script moving fwd tho.

.github/workflows/publish-packages.yaml
DanRod1999: Can we rename the workflow with old or archived or something and then just disable it?

DanRod1999: nvm renaming is weird, and idk how disabling one of them would work in this case. Workflows are a bit strange in comparison to ado pipelines it seems.

.github/workflows/scripts/create-release.sh
anmolshres98: all in all I think shell scripting language decision have to be a case by case basis. It would be harder to mandate and enforce a specific technology or methodology, specially with the ever changing landscape of js libraries and programming paradigms. Personally, [Occam's razor](https://en.wikipedia.org/wiki/Occam%27s_razor)ing my way through most shell scripting usecases leads me to lean more towards native scripting

.github/workflows/scripts/create-release.sh
anmolshres98: yes this would zip up all of and just the source code. My implementation is aiming to go for the most recent working version of when releases page was working. In that version the entire repo's code is made available for every package release. Hence, I tried to narrow it to just the src code of the released package. I do now realize that the flag to specify assets is so as to release binaries or libs and not the src code itself. So, I could do one of 2 things:

- go back to the way it was so that we don't zip anything manually and don't specify any assets. That way the src code for entire repo is available for download by default
- pnpm install repo then build that specific package and zip the newly created build artifacts as assets instead

lmk what you think would be best in our case

.github/workflows/publish-packages.yaml
aruniverse: we can disable workflows

anmolshres98: perhaps the best option is to instead of having this release be created in a separate workflow, we do it all in the current "Create Release" workflow. We could rename it to something more broad. This way the release is created the same time as the tags are. Since the npm publishing is tightly coupled (imo) to creating github releases already, this would make logical sense. Also, less compute used for spinning up new vm for small work (very small cost probably but still). This would also retain the history in all one location. We then delete/disable the newly created pipelines that have not done any "production" work so far so no harm no foul.

What do you guys think?

anmolshres98: it seems the future has yet [not arrived](https://github.com/orgs/community/discussions/26256#discussioncomment-3250998) but there seems to be [workaround for deleting workflows](https://github.com/orgs/community/discussions/26256#discussioncomment-6113022)

Since, we don't have any important history to be saved in the newly created workflows, we could delete both

anmolshres98: Actually, we would have the added complexity of figuring out which package actually had updates and newly created tags if we use the same workflow to do both but having them separated gives us the benefit to not have to figure that out since the a new run is triggered for each tag separately.

aruniverse: lets keep this as is, ill just disable the old workflow

.github/workflows/scripts/create-release.sh
ben-polinsky: In addition, it seems [github already does this](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#about-releases): 

> GitHub will automatically include links to download a zip file and a tarball containing the contents of the repository at the point of the tag's creation.

anmolshres98: Yes, I agree guys. Yes, the zipping of entire src code is done by github on default. I was simply explaining my thought process as to what lead me to zipping the package specific src code. Only recently I realized that we should be zipping the importable libs and bins instead. 

I will update so that that is being done correctly, thanks.

.github/workflows/publish-packages.yaml
```diff
@@ -18,19 +16,13 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v4
 
-      - name: Use Node.js 18
-        uses: actions/setup-node@v3
-        with:
-          node-version: 18.x
-          registry-url: https://registry.npmjs.org/
-
```
anmolshres98, line 26: note needed:

https://github.com/iTwin/auth-clients/actions/runs/12837764292/job/35802083363

.github/workflows/scripts/create-release.sh
anmolshres98: Addressed the code so now it we are zipping the deliverables correctly:

https://github.com/iTwin/auth-clients/releases/tag/%40itwin%2Felectron-authorization_v0.19.8

PR #285, opened 2025-02-17
Reintroduce actions/setup-node and make sure registry-url is provided for npm publish authentication

hl662:
From this [SO](https://stackoverflow.com/questions/75766122/whats-the-difference-between-node-auth-token-and-npm-auth-token#comment138206931_77539464) post, we can derive that we need to specify the `registry-url` when setting up node. That way, `NODE_AUTH_TOKEN` would be used and added to `.npmrc` during our CI runs.

Also, we have to re-add `actions/setup-node` to our yaml scripts, because `pnpm/action-setup` [does not replace it](https://github.com/pnpm/action-setup?tab=readme-ov-file#notes). Also bumped the setup-node actions in `ci.yaml` to use v4 alongside.

Dry Run success: https://github.com/iTwin/auth-clients/actions/runs/13378305641/job/37362067098

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,9 +28,9 @@ jobs:
           git config --local user.name imodeljs-admin
 
       - name: Setup node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
-          node-version: 18.x
+          node-version: 20.x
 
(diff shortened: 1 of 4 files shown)


ben-polinsky (approved):

aruniverse (approved):
lgtm

PR #287, opened 2025-03-20
Updating `open` package from 8.3.0 to 10.1.0

wolfo951:
The old 8.3.0 version does not work on all machines and contains a problem. The latest version solves those problems.


packages/node-cli/src/Client.ts
aruniverse, line 1: We're there actual changes to this file outside of whitespace?
Can you revert the whitespace changes to the file? You probably have a formatter enabled in your editor.

aruniverse (commented):
Please also add change file

packages/node-cli/src/Client.ts
wolfo951: Yes, there are changes outside of whitespace. I have reverted all the formatting changed that was done by my editor.

packages/node-cli/src/Client.ts
```diff
@@ -443,7 +443,8 @@ class NodeCliAuthorizationRequestHandler extends AuthorizationRequestHandler {
 
     // Compose the request and invoke in the browser
     const authUrl = this.buildRequestUrl(serviceConfiguration, authRequest);
-    await open(authUrl);
+    const open = await import("open");
```
aruniverse, line 446: we already import `open` at the top

wolfo951: I have removed import at the top because the latest version of `open` is native ESM and it cannot be imported like that anymore.

aruniverse:
```ts
import open from "open"
```

this doesnt work?

wolfo951: No. I get an error:
```
The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("open")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to 'c:/Development/auth-clients/packages/node-cli/package.json'
```

aruniverse: i think there is a tsconfig option to allow for esModuleInterop which should allow for the above change

wolfo951: I tried to add `"esModuleInterop": true` to the `tsconfig.json` But then I had to update tests. I updated the test configuration but there was still a failure in the setup. Did not tried to fix this further on as it already needed a lot of module changes. 

Is there a reason why suggestions solution form the IDE does not work in this project?

aruniverse: what was the suggested solution from the ide? is it what you have rn?

wolfo951: Yes. This solution was suggestion from the error message in the IDE (VS Code)

ben-polinsky:
Linux integration tests appear flaky on `main` as well, I wouldn't hold this because of them. Issue raised: https://github.com/iTwin/auth-clients/issues/298

aruniverse (approved):

ben-polinsky (approved):

aruniverse:
@wolfo951 can you resolve the conflict, we have addressed the integration test failures in master already

PR #288, opened 2025-03-25
Add missing '@itwin/core-geometry' dependency to service package

GytisCepk:
Because of missing peer dependency, users may see:
```
 WARN  Issues with peer dependencies found
 ...
  └─┬ @itwin/service-authorization 1.2.2
    └─┬ @itwin/core-common 4.2.1
      └── ✕ missing peer @itwin/core-geometry@^4.2.1
Peer dependencies that should be installed:
  @itwin/core-geometry@^4.2.1
```

### DIFF
--- a/packages/service/package.json
+++ b/packages/service/package.json
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@itwin/core-common": "^3.3.0 || ^4.0.0",
+    "@itwin/core-geometry": "^3.3.0 || ^4.0.0",
     "got": "^12.6.1",
     "jsonwebtoken": "^9.0.0",
     "jwks-rsa": "^2.0.4"
--- a/change/@itwin-service-authorization-25a3c167-db19-4a67-97fd-77e7a9d5d396.json
+++ b/change/@itwin-service-authorization-25a3c167-db19-4a67-97fd-77e7a9d5d396.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add '@itwin/core-geometry' as a peer dep to service package",
+  "packageName": "@itwin/service-authorization",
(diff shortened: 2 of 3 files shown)


packages/service/package.json
```diff
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@itwin/core-common": "^3.3.0 || ^4.0.0",
+    "@itwin/core-geometry": "^3.3.0 || ^4.0.0",
```
aruniverse, line 31: could it be just a dev dep?

GytisCepk: I don't believe so. Aren't package dev dependencies ignored when said packages is installed as dependency ?

aruniverse: Well, we dont directly use anything in core-geom, we just need to satisfy the peer of core-common, so it could suffice as a dev dep i think. If consumers are running into an issue with a missing peer on core-geom, they could add it on their end, but shouldnt need to.

saskliutas: I don't think adding it to dev deps is the right solution. It should be a direct dependency because this package is a consumer of `@itwin/core-common` and should fulfill it's contract which at the moment requires that `@itwin/core-geometry` should be installed as well. If it's weird that you need to have `core-geometry` dep in order to us `core-common` it should be solved at `@itwin/core-common` package instead of work arounding it in here. Maybe `core-geometry` should be an optional peer dep if there is valid use case for using `core-common`without `core-geometry`.

This is also highlighted by your comment: `If consumers are running into an issue with a missing peer on core-geom, they could add it on their end` as this package does not state anywhere that it might not work if you don't have `core-geometry` as dep.

ben-polinsky (approved):

PR #300, opened 2025-05-29
Add support for Electron 34, 35 and 36

GytisCepk:
Even though auth packages do not support iTwin.js 5.0 rc officially, there are products which try to use it (e.g. https://github.com/iTwin/viewer/pull/366). This is a problem since there is no overlap in supported Electron versions between 4.x and 5.0 rc.

Once auth packages drop iTwin.js 4.x, support, we should drop support for old Electron versions as well.


aruniverse (approved):

aruniverse:
cc https://github.com/iTwin/auth-clients/pull/301#issuecomment-2920581862

ben-polinsky (approved):

PR #301, opened 2025-05-29
Add support for core 5.0.0

pankhur94:
Updating to support core 5.0 since this is causing core-* mismatches in preparing of next version for our viewer.

### DIFF
--- a/packages/electron/src/test/MainAuthorizationClient.test.ts
+++ b/packages/electron/src/test/MainAuthorizationClient.test.ts
@@ -12,7 +12,7 @@ import { ElectronMainAuthorization } from "../main/Client";
 import { RefreshTokenStore } from "../main/TokenStore";
 import { getConfig, getMockTokenResponse, setupMockAuthServer, stubTokenCrypto } from "./helpers/testHelper";
 /* eslint-disable @typescript-eslint/naming-convention */
-const assert = chai.assert;
+const assert: Chai.AssertStatic = chai.assert;
 const expect = chai.expect;
 
 chai.use(chaiAsPromised);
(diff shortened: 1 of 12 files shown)


packages/electron/src/test/MainAuthorizationClient.test.ts
```diff
@@ -12,7 +12,7 @@ import { ElectronMainAuthorization } from "../main/Client";
 import { RefreshTokenStore } from "../main/TokenStore";
 import { getConfig, getMockTokenResponse, setupMockAuthServer, stubTokenCrypto } from "./helpers/testHelper";
 /* eslint-disable @typescript-eslint/naming-convention */
-const assert = chai.assert;
+const assert: Chai.AssertStatic = chai.assert;
```
aruniverse, line 15: whats the need for the explicit typing?

you could probably do the following

```ts
const { assert, expect } = chai;
```

aruniverse (approved):

MichaelSwigerAtBentley (approved):

aruniverse:
I initially approved this thinking we could avoid a major version bump; but after reading https://github.com/iTwin/viewer/pull/366 i think we should just bite the bullet and clean up our deps and do a major

- lets drop support for 3.x and 4.x
- and move the direct dep on core-common to a dev dep snce its only used as types
- for browser-auth, lets bump oidc-client-ts to the the latest major
- for electron-auth, lets drop older versions of electron like @GytisCepk mentioned in https://github.com/iTwin/auth-clients/pull/300
- for service, we should be able to drop dep on core-geom, and update jwks-rsa to latest major

maybe we split some of these into separate pr, targeting different pkgs

pankhur94:
closed in favour of https://github.com/iTwin/auth-clients/pull/305 and https://github.com/iTwin/auth-clients/pull/306.

PR #312, opened 2025-07-11
Add support for Electron 37

GytisCepk:
[Electron 37 breaking changes](https://github.com/electron/electron/blob/main/docs/breaking-changes.md#planned-breaking-api-changes-370)

@saskliutas, bringing balance to the force

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -31,13 +31,13 @@
     "directory": "packages/browser"
   },
   "dependencies": {
-    "@itwin/core-common": "^4.11.6",
+    "@itwin/core-common": "^5.0.0",
     "oidc-client-ts": "^3.3.0"
   },
   "devDependencies": {
-    "@itwin/build-tools": "rc",
-    "@itwin/core-bentley": "rc",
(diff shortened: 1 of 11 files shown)


packages/browser/package.json
```diff
@@ -31,13 +31,13 @@
     "directory": "packages/browser"
   },
   "dependencies": {
-    "@itwin/core-common": "^4.11.6",
+    "@itwin/core-common": "^5.0.0",
```
saskliutas, line 34: (Unrelated to these changes resolve any time)

This part looks weird. This package has direct dependency on `core-common` and peer dep on `core-bentley`. At the same time `core-common` has peer dep on `core-bentley` as well. This means that `core-common` and `core-bentley` versions resolved when installing this package need to be the same. However, `core-bentley` version is controlled by consumer. Shouldn't this make `core-common` also a peer dependency? Or do we trust that package manager will try to match `core-common` version to `core-bentley` we received from consumer?

GytisCepk:
@ben-polinsky @anmolshres98 @MichaelSwigerAtBentley, please review

PR #314, opened 2025-07-21
Add the ability to send additional request headers in service authorization client

paulius-valiunas:
This can be used to send headers such as `X-Correlation-Id` to the authorization server

### DIFF
--- a/packages/service/src/ServiceAuthorizationClient.ts
+++ b/packages/service/src/ServiceAuthorizationClient.ts
@@ -51,12 +51,12 @@ export class ServiceAuthorizationClient implements AuthorizationClient {
     this._configuration = serviceConfiguration;
   }
 
-  private async generateAccessToken(): Promise<string> {
+  private async generateAccessToken(additionalHeaders?: { [key: string]: string }): Promise<string> {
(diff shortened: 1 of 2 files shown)


ben-polinsky:
Cool. Could you add a note in the readme?

packages/service/src/ServiceAuthorizationClient.ts
```diff
@@ -112,9 +113,9 @@ export class ServiceAuthorizationClient implements AuthorizationClient {
   /** Returns a promise that resolves to the AccessToken of the currently authorized client.
   * The token is refreshed if necessary.
   */
-  public async getAccessToken(): Promise<string> {
+  public async getAccessToken(additionalHeaders?: { [key: string]: string }): Promise<string> {
```
MichaelSwigerAtBentley, line 116: Could we add a JS doc comment clarifying the additionalHeaders

paulius-valiunas: I've added both a jsdoc tag and an explanation in the readme. I'm just wondering, do we really want to keep duplicating our documentation between jsdoc and readme?

PR #317, opened 2025-08-25
move testing deps to devDependencies

kckst8:
Not sure if there was a reason to keep in dependencies. The problem that iTwin Studio ran into was when installing deps for a production build the certa postinstall was executing, which attempts to download and install chromium from playwright. This is quite costly

### DIFF
--- a/packages/oidc-signin-tool/package.json
+++ b/packages/oidc-signin-tool/package.json
@@ -33,18 +33,18 @@
     "url": "http://www.bentley.com"
   },
   "dependencies": {
-    "@itwin/certa": "^5.0.0",
     "@itwin/service-authorization": "workspace:^",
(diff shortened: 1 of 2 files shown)


MichaelSwigerAtBentley (approved):

ben-polinsky:
> The problem that iTwin Studio ran into was when installing deps for a production build the certa postinstall was executing

Shouldn't this package (sign in tool) be dev only for studio?

ben-polinsky:
The signin-tool uses certa at runtime...

anmolshres98:
closing since the assessment from https://github.com/iTwin/auth-clients/pull/317#issuecomment-3221031247 is correct imo. studio should not be bundling test dependencies into prod builds

PR #326, opened 2025-12-23
Allow customizing IPC channel names

alexdunae:
Apps may need to have more than one instance the Electron auth client with the same `clientId`.  In these cases, the IPC channel names will conflict.

This PR adds the ability to make the IPC channel names unique per-client.


ben-polinsky (approved):

alexdunae:
Replaced by #328

PR #332, opened 2026-01-27
Fix ReLogin in E2E due to missing locator

bilalbaig3101:
This PR fixes an issue in the E2E tests where relogin fails. On reLogin the function is unable to locate the email input field causing the test to fail. 

**Root Cause**

After the signout the login page shows the already loged out user email instead of email input field which is something login funciton relies on 
_this is how the login page is showen when cache is cleared_
Screenshot 2026-01-26 191324
_this is how the login page on re-login and the email locator is missing_
Screenshot 2026-01-27 145345

**Solution**
added a locator to click "Use another accout"


bilalbaig3101:
> Thanks for the contribution. E2E tests are failing, however. Have you thoroughly tested this?

@ben-polinsky re-ran the pipeline all test stages passed this time

ben-polinsky (approved):

PR #345, opened 2026-04-17
feat!: add @itwin/service-authorization as a peer dependency of @itwin/oidc-signin-tool

### LINKED ISSUE #344: oidc-signin-tool has a direct dependency on a package from this monorepo, it should be a peerdependency
See [package.json](https://github.com/iTwin/auth-clients/blob/main/packages/oidc-signin-tool/package.json)

Notice that it has a direct dependency on:
    "@itwin/service-authorization": "workspace:^",

This should be a peepDependency ^2.1.0 and a devDependency of "workspace:^"

Otherwise, it forces all consuming apps to be on the same version of @itwin/service-authorization and leads to doppelgangers with other iTwin or Bentley packages.

I believe that are plenty examples of this in other iTwin package.json

ben-polinsky:
## Why

Closes #344.

\`@itwin/oidc-signin-tool\` was declaring \`@itwin/service-authorization\` as a direct \`dependency\`. Because pnpm creates separate physical installs per unique peer set, this caused apps that also depended on \`service-authorization\` (at a different version) to end up with two copies in the tree. Any package that registers singletons at module load time (like \`civil-backend\`) would break at runtime — two copies means two registries.

Moving it to \`peerDependencies\` removes \`oidc-signin-tool\` as an independent version source and lets the consuming app own a single resolved copy.

## What

- \`@itwin/service-authorization\` moved from \`dependencies\` → \`peerDependencies: "^2.1.0"\` in \`@itwin/oidc-signin-tool\`
- Added to \`devDependencies: workspace:^\` so local tests still work
- **Breaking change** — consumers must now explicitly declare \`@itwin/service-authorization\` in their own \`dependencies\` or \`devDependencies\`


aruniverse:
@paulius-valiunas @saskliutas can you guys review?

packages/oidc-signin-tool/package.json
```diff
@@ -61,6 +61,7 @@
     "typescript": "~5.6.3"
   },
   "peerDependencies": {
-    "@itwin/core-bentley": "^5.0.0"
+    "@itwin/core-bentley": "^5.0.0",
+    "@itwin/service-authorization": "^2.1.0"
```
aruniverse, line 65: should this be using the workspace protocol?

ben-polinsky: yes, thanks

ben-polinsky: Actually, no, from what I understand it will hard code the full version on publish, restricting consumers to 2.1.1 or whatever we have here.

anmolshres98:
Relates to https://github.com/iTwin/auth-clients/issues/344

packages/oidc-signin-tool/package.json
anmolshres98: `workspace:^` should work https://pnpm.io/workspaces#publishing-workspace-packages

saskliutas: I don't think `workspace:^` works for specifying peer dep version. From my experience when `@itwin/service-authorization` is bumped to `2.2.0` or `2.1.1` during publish this peer will resolve to `^2.2.0` or `^2.1.1` and bumping peer dep is major change.

I think hard coded version should be used. @ben-polinsky I would also recommend to try running bump locally to check if beachball does not change this. I remember workspace packages even without `workspace:` protocol were bumped.

saskliutas (approved):
LGTM

aruniverse (approved):

ben-polinsky:
We're not sure this is the correct fix or necessary. See https://github.com/iTwin/auth-clients/issues/344

hl662:
@ben-polinsky 

I've marked the PR active. We're moving ahead with this

PR #348, opened 2026-06-18
fix: resolve critical/high pnpm audit vulnerabilities + prune stale overrides

hl662:
## Why

> Making audit resolution separate from https://github.com/iTwin/auth-clients/pull/345

Resolves critical/high CVEs found by \`pnpm audit\` as of 2026-06-18. Also aligns all CI pipelines to Node 24 LTS while here — pipelines were on deprecated \`NodeTool@0\` / Node 20.

Resolve some critical and high CVEs, and clean up some stale ones.

### Pipeline fixes
- Replace deprecated \`NodeTool@0\` (Node 20) with \`UseNode@1\` pinned to \`24.15.0\` across all ADO + GH Actions pipelines (\`build-docs.yaml\`, \`integration-test.yaml\`, \`ci.yaml\`, \`publish-packages.yaml\`, \`create-github-release.yaml\`)
- Remove \`ignoreAudit: true\` from \`build-docs.yaml\` — parameter no longer exists in the itwinjs-core \`docs-build.yaml\` template, was causing YAML validation failure on every PR build

### Direct dependency bump
\`vite\` is a real devDependency of \`@itwin/browser-authorization\` and \`@itwin/electron-authorization\`, so it's bumped at the source rather than forced through an override:
- \`vite\` \`^6.4.2\` → \`^6.4.3\` — fixes \`server.fs.deny\` bypass (high) + launch-editor NTLMv2 disclosure (moderate)

### Scoped \`pnpm.overrides\` (deep transitives we don't declare)
| Severity | Package | Patched to |
|---|---|---|
| critical | shell-quote | \`>=1.8.4\` |
| high | fast-uri | \`3.1.2\` (pinned exact — \`>=3.1.2\` would resolve past ajv's \`^3.0.1\` into 4.x) |
| high | form-data | \`>=4.0.6\` |
| high | undici | \`>=7.28.0 <8\` |

\`undici@7.25.0\` comes in via \`@electron/get@5.0.0\`; the override is constrained to \`<8\` to stay within \`@electron/get\`'s expected major.

Override keys are version-scoped so non-vulnerable copies elsewhere in the tree are untouched.

### Stale overrides removed
Both no longer match any vulnerable resolution and were verified safe to drop:
- \`path-to-regexp@<0.1.13\` — sole consumer \`express@4.22.1\` now declares \`~0.1.12\`, resolving \`0.1.13\` (patched) on its own
- \`picomatch@<2.3.2\` — nothing in the tree resolves below \`2.3.2\`; dependents declare \`^2.3.1\` / \`^3 || ^4\`

🤖 Nambot (Opus 4.8) and Nam edits


package.json
```diff
@@ -39,10 +39,12 @@
...
   "pnpm": {
     "overrides": {
-      "path-to-regexp@<0.1.13": "0.1.13",
       "lodash@<=4.17.23": "4.18.1",
       "serialize-javascript@<=7.0.2": ">=7.0.3",
-      "picomatch@<2.3.2": ">=2.3.2"
+      "shell-quote@>=1.1.0 <=1.8.3": ">=1.8.4",
+      "fast-uri@<3.1.2": ">=3.1.2",
+      "form-data@>=4.0.0 <4.0.6": ">=4.0.6",
+      "undici@>=7.0.0 <7.28.0": ">=7.28.0 <8"
```
anmolshres98, line 47: woof, no other way around it? `update -r` ?

aruniverse (approved):

package.json
hl662: Alrighty just checked: update -r does resolve all four; the patched versions are in-range, so they aren't hard-pinned. But uh, there's a blast radius: it would re-resolve the whole workspace (~1.4k lockfile lines here), bumps unrelated deps like @itwin/core-common to 5.10.3, and that bump introduces a fresh core-common/core-geometry peer mismatch (5.10.3 vs 5.7.3) we'd then have to sort out — all to fix four leaf transitives.

Should we do that recursive bump in here? Sounds like the equivalent of solving our rush update --full problem in core earlier this year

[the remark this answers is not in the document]
hl662: Keeping the range intentionally — lockfile already pins 7.28.0, the <8 cap prevents major bumps, and future 7.x bug fixes would be welcome on regenerate. Pinning exact would require a manual override update for any future 7.x security patches.

Nambot 🤖 (powered by claude-sonnet-4-6)

anmolshres98: nah this is fine

anmolshres98 (approved):

hl662:
excuse me, just doing some boy scout and bumping pipelines to node 24 and non deprecated usenode task

PR #350, opened 2026-06-26
Allow passing in `playwright` page

### LINKED ISSUE #349: Allow passing in already loaded `playwright` instance
Direct dependency on `@playwright/test` causes issues when `@itwin/oidc-signin-tool` is used in e2e tests. We need to use same `@playwright/test` version or override `@itwin/oidc-signin-tool` to use our version of `@playwright/test`. If multiple versions of `@playwright/test` are present in node_modules `@itwin/oidc-signin-tool` workflow fails with playwright error:

`Error: Could not load @playwright/test. Do you have multiple playwright dependencies active? If so, then you should provide your own playwright Page to automation APIs to avoid us attempting to make our own by importing playwright`

saskliutas:
Closes https://github.com/iTwin/auth-clients/issues/349

Added optional `page` parameter to the methods doing signing. This allows to pass in already loaded `playwright` instance and avoid `oidc-signin-tool` loading its own.

Tested manually in e2e tests and this allows to use different playwright version without overriding `oidc-signin-tool` playwright version. ~~One caveat is that there might by typescript errors when using newer version than one specified `oidc-signin-tool`. This is due `playwright` not following semver and making breaking changes in minor bump. However, everything worked fine as not whole API surface is used by `oidc-signin-tool`.~~ I could successfully pass in Page from different versions of playwright to be used for signin process.

### DIFF
--- a/change/@itwin-oidc-signin-tool-f94d62ff-891e-4f65-a154-dd105ce33a9b.json
+++ b/change/@itwin-oidc-signin-tool-f94d62ff-891e-4f65-a154-dd105ce33a9b.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add optional page parameter to getAccessToken/getTestAccessToken/TestUtility.getAccessToken so consumers can supply their own Playwright Page, avoiding 'multiple playwright dependencies active' runtime errors.",
+  "packageName": "@itwin/oidc-signin-tool",
+  "email": "24278440+saskliutas@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@itwin-oidc-signin-tool-f94d62ff-891e-4f65-a154-dd105ce33a9b.json
+++ b/change/@itwin-oidc-signin-tool-f94d62ff-891e-4f65-a154-dd105ce33a9b.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Add optional page parameter to getAccessToken/getTestAccessToken/TestUtility.getAccessToken so consumers can supply their own Playwright Page, avoiding 'multiple playwright dependencies active' runtime errors.",
+  "comment": "Add optional Playwright `page` parameter to getAccessToken/getTestAccessToken/TestUtility.getAccessToken so consumers can supply their own Page.",
   "packageName": "@itwin/oidc-signin-tool",
   "email": "24278440+saskliutas@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/packages/oidc-signin-tool/README.md
+++ b/packages/oidc-signin-tool/README.md
@@ -50,6 +50,42 @@ client.hasSignedIn
 
 ```
 
+## Bring your own Playwright
+
+By default the tool dynamically imports its own copy of `@playwright/test` to
+launch the headless browser that drives the sign-in UI. If your project already
+depends on Playwright, loading two Playwright instances at once throws:
+
+> Could not load @playwright/test. Do you have multiple playwright dependencies active? ...
+
+To avoid this, build a Playwright `Page` with your own Playwright and pass it via
+the `page` option. The tool then drives that page instead of importing its own
+Playwright, so no second instance is loaded:
+
+```typescript
+import { chromium } from "@playwright/test";
+import { getTestAccessToken, TestBrowserAuthorizationClient } from "@itwin/oidc-signin-tool";
+
+const browser = await chromium.launch();
+const page = await browser.newPage();
+
+// convenience function
+const accessToken = await getTestAccessToken(config, user, { page });
+
+// or on the client
+const client = new TestBrowserAuthorizationClient(config, user);
+const token = await client.getAccessToken({ page });
+
+// you own the lifecycle of the page and browser:
+await browser.close();
+```
+
+The option is also available on `client.signIn()` and
+`TestUtility.getAccessToken()`. When you supply a `page`, the tool does **not**
+close your page or browser — you are responsible for cleaning them up. When the
+option is omitted, the existing dynamic-import behavior is used, so this is fully
+backwards compatible.
+
 ## Mocha leaks
 
 If you use this package with mocha, you may notice a global leak.
@@ -65,6 +101,9 @@ used to ignore because it was created during the import phase. Now it is created
 conditional import where mocha doesn't expect it and it "leaks". **You should configure mocha to ignore
 this leak.** In the future, the API requiring the dynamic import will be moved to a separate package
 to avoid this situation.
+
+> Note: passing your own `page` (see [Bring your own Playwright](#bring-your-own-playwright))
+> skips the dynamic import entirely, which also avoids this leak.
 
 You can ignore the leak like so:
 
--- a/packages/oidc-signin-tool/README.md
+++ b/packages/oidc-signin-tool/README.md
@@ -83,8 +83,7 @@ await browser.close();
 The option is also available on `client.signIn()` and
 `TestUtility.getAccessToken()`. When you supply a `page`, the tool does **not**
 close your page or browser — you are responsible for cleaning them up. When the
-option is omitted, the existing dynamic-import behavior is used, so this is fully
-backwards compatible.
+option is omitted, the default dynamic import behavior is used.
 
 ## Mocha leaks
 
--- a/packages/oidc-signin-tool/src/SignInAutomation.ts
+++ b/packages/oidc-signin-tool/src/SignInAutomation.ts
@@ -33,6 +33,10 @@ interface AutomatedContextBase<T> {
 
   /** whether or not to kill the entire browser when cleaning up */
   doNotKillBrowser?: boolean;
+
+  /** whether or not to leave the page open when cleaning up.
+   * Useful when the consumer supplies their own page and owns its lifecycle. */
+  doNotClosePage?: boolean;
 }
 
 /** @internal context for automated sign in functions */
@@ -87,7 +91,7 @@ export async function automatedSignIn<T>(
       // eslint-disable-next-line @typescript-eslint/return-await
       return await context.resultFromCallback(await waitForCallback);
   } finally {
-    await cleanup(page, controller.signal, waitForCallback, context.doNotKillBrowser);
+    await cleanup(page, controller.signal, waitForCallback, context.doNotKillBrowser, context.doNotClosePage);
   }
 }
 
@@ -106,7 +110,7 @@ export async function automatedSignOut<T>(
   try {
     await page.goto(context.signOutInitUrl);
   } finally {
-    await cleanup(page, controller.signal, waitForCallback, context.doNotKillBrowser);
+    await cleanup(page, controller.signal, waitForCallback, context.doNotKillBrowser, context.doNotClosePage);
   }
 }
 
@@ -351,10 +355,15 @@ async function cleanup(
   signal: AbortSignal,
   waitForCallbackUrl: Promise<any>,
   doNotKillBrowser = false,
+  doNotClosePage = false,
 ) {
   if (signal.aborted)
     await page.reload();
   await waitForCallbackUrl;
+
+  if (doNotClosePage)
+    return;
+
   await page.close();
 
   const doKillBrowser = !doNotKillBrowser;
--- a/packages/oidc-signin-tool/src/SignInAutomation.ts
+++ b/packages/oidc-signin-tool/src/SignInAutomation.ts
@@ -8,6 +8,77 @@ import type { Browser, LaunchOptions, Page } from "@playwright/test";
 import type { TestUserCredentials } from "./TestUsers";
 import { testSelectors } from "./TestSelectors";
 
+/* The following `*Like` interfaces describe the minimal subset of the Playwright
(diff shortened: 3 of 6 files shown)


aruniverse (approved):

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
```diff
@@ -6,13 +6,35 @@ import type { AccessToken } from "@itwin/core-bentley";
...
+   * already depend on Playwright to avoid loading two Playwright instances,
+   * which throws a "multiple playwright dependencies active" runtime error.
+   *
+   * The tool does **not** close your page or browser — you own their lifecycle.
+   *
+   * When omitted, the tool falls back to importing its own `@playwright/test`
+   * and launching its own browser.
+   */
+  page?: Page;
+}
```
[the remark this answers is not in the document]
saskliutas: Asked copilot to extract PageLike types that would cover API surface used for signin automation. Now Page from different playwright versions could be passed in without type errors.

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
```diff
@@ -128,7 +150,8 @@ export class TestBrowserAuthorizationClient implements AuthorizationClient {
 
     const controller = new AbortController();
 
-    const page = await SignInAutomation.launchDefaultAutomationPage();
+    const consumerOwnsPage = options?.page !== undefined;
+    const page = options?.page ?? await SignInAutomation.launchDefaultAutomationPage();
```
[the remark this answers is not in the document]
saskliutas: Added tests

saskliutas:
@hl662 @ben-polinsky could you take a look or am I good to merge this?

change/@itwin-oidc-signin-tool-f94d62ff-891e-4f65-a154-dd105ce33a9b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add optional page parameter to getAccessToken/getTestAccessToken/TestUtility.getAccessToken so consumers can supply their own Playwright Page, avoiding 'multiple playwright dependencies active' runtime errors.",
```
anmolshres98: this could be shortened to just the new feature that was added

packages/oidc-signin-tool/src/SignInAutomation.ts
```diff
@@ -8,6 +8,77 @@ import type { Browser, LaunchOptions, Page } from "@playwright/test";
...
+export interface RequestLike {
+  url(): string;
+}
+
+/**
+ * The minimal subset of the Playwright `Page` API used by the automation.
+ *
+ * A real Playwright `Page` satisfies this interface, so you can supply a page
+ * created by *your own* Playwright installation/version without being forced
+ * onto this package's exact `@playwright/test` types.
```
anmolshres98: is this portion needed? could line 61 be enough? like the other interfaces you added in this file? mostly I am not a fan of the "*your own*" portion

packages/oidc-signin-tool/src/SignInAutomation.ts
```diff
@@ -33,6 +104,10 @@ interface AutomatedContextBase<T> {
 
   /** whether or not to kill the entire browser when cleaning up */
   doNotKillBrowser?: boolean;
+
+  /** whether or not to leave the page open when cleaning up.
+   * Useful when the consumer supplies their own page and owns its lifecycle. */
+  doNotClosePage?: boolean;
```
anmolshres98:
```suggestion
  closePage?: boolean;
```
this is easier to read imo

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
```diff
@@ -12,6 +12,33 @@ import type {
...
+   *
+   * The tool does **not** close your page or browser — you own their lifecycle.
+   *
+   * When omitted, the tool falls back to importing its own `@playwright/test`
+   * and launching its own browser.
+   *
+   * Typed as the minimal structural {@link PageLike} rather than Playwright's
+   * concrete `Page` so a page from any Playwright installation/version is
+   * accepted without forcing consumers onto this package's `@playwright/test`
+   * types. A real Playwright `Page` satisfies {@link PageLike}.
```
anmolshres98:
```suggestion
```
I don't think it's necessary

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
```diff
@@ -12,6 +12,33 @@ import type {
...
+export interface GetAccessTokenOptions {
+  /**
+   * A Playwright `Page` used to drive the sign-in UI.
+   *
+   * When provided, the tool uses this page instead of dynamically importing its
+   * own copy of `@playwright/test` and launching a browser. Supply this when you
+   * already depend on Playwright to avoid loading two Playwright instances,
+   * which throws a "multiple playwright dependencies active" runtime error.
+   *
+   * The tool does **not** close your page or browser — you own their lifecycle.
```
anmolshres98:
```suggestion
   * The tool does not close the page or browser, caller is responsible for the page's lifecycle.
```

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
```diff
@@ -12,6 +12,33 @@ import type {
...
+   *
+   * When omitted, the tool falls back to importing its own `@playwright/test`
+   * and launching its own browser.
```
anmolshres98:
```suggestion
```
this seems redundant info since line 25-26 already implies this imo

packages/oidc-signin-tool/src/TestBrowserAuthorizationClient.ts
```diff
@@ -12,6 +12,33 @@ import type {
...
+ * @alpha
+ */
+export interface GetAccessTokenOptions {
+  /**
+   * A Playwright `Page` used to drive the sign-in UI.
+   *
+   * When provided, the tool uses this page instead of dynamically importing its
+   * own copy of `@playwright/test` and launching a browser. Supply this when you
+   * already depend on Playwright to avoid loading two Playwright instances,
+   * which throws a "multiple playwright dependencies active" runtime error.
```
anmolshres98:
```suggestion
   * already depend on Playwright to avoid loading two Playwright instances.
```

packages/oidc-signin-tool/src/TestUtility.ts
```diff
@@ -47,10 +48,12 @@ export class TestUtility {
    * to signin the user through a headless browser.
    * - Uses the default iModel.js internal OIDC SPA client registration
    * @param user Test user credentials
+   * @param options Optionally supply your own Playwright `page` to avoid
```
anmolshres98:
```suggestion
   * @param options Optionally supply your a Playwright `page` to avoid
```

packages/oidc-signin-tool/README.md
```diff
@@ -50,6 +50,42 @@ client.hasSignedIn
...
+
+// you own the lifecycle of the page and browser:
+await browser.close();
+```
+
+The option is also available on `client.signIn()` and
+`TestUtility.getAccessToken()`. When you supply a `page`, the tool does **not**
+close your page or browser — you are responsible for cleaning them up. When the
+option is omitted, the existing dynamic-import behavior is used, so this is fully
+backwards compatible.
```
anmolshres98:
```suggestion
The option is also available on `client.signIn()` and
`TestUtility.getAccessToken()`. When you supply a `page`, the tool does **not**
close your page or browser — you are responsible for cleaning them up. When the
option is omitted, the default dynamic import behavior is used.
```
This document is not a changelog and instead could be read by someone who is using this tool for the first time. In that context, there shouldn't be concepts of "existing" behavior or "backwards-compatibilit"

packages/oidc-signin-tool/src/SignInAutomation.ts
```diff
@@ -8,6 +8,77 @@ import type { Browser, LaunchOptions, Page } from "@playwright/test";
...
+ *
+ * Return types for calls whose results are ignored are widened to
+ * `Promise<unknown>` so that Playwright's more specific return types (e.g.
+ * `Promise<Response | null>`) remain assignable. */
+
+/**
+ * The minimal subset of the Playwright `Locator` API used by the automation.
+ * @alpha
+ */
+export interface LocatorLike {
```
anmolshres98, line 21: Is it possible to make `Promise<unknown>` narrower where return types are predictable?

Example:

LocatorLike.waitFor(...) → Promise<void> (instead of Promise<unknown>)
Likewise fill/click can be Promise<void>, and goto/reload can be a lightweight ResponseLike | null if needed.

Promise<unknown> is safe but too broad imo. feels like we could reasonably guess a narrower type?

anmolshres98 (commented):
overall looks good, quite useful thanks !

packages/oidc-signin-tool/src/SignInAutomation.ts
```diff
@@ -8,6 +8,77 @@ import type { Browser, LaunchOptions, Page } from "@playwright/test";
...
+ * without baking this package's exact `@playwright/test` type identity into the
+ * public API. A real Playwright `Page` is assignable to `PageLike`.
+ *
+ * Return types for calls whose results are ignored are widened to
+ * `Promise<unknown>` so that Playwright's more specific return types (e.g.
+ * `Promise<Response | null>`) remain assignable. */
+
+/**
+ * The minimal subset of the Playwright `Locator` API used by the automation.
+ * @alpha
```
anmolshres98, line 19: should all these changes be `@beta`? @hl662 thoughts?

ben-polinsky (approved):
LGTM, thanks

packages/oidc-signin-tool/src/SignInAutomation.ts
hl662: If it was an API we intend for our app consumers to use then I would say so, especially considering the apis are well documented.

But... it's a dev tooling api, and we are hitching ourselves to a dependency that we do not maintain/control/ can contribute quick enough so I would rather keep it as alpha

packages/oidc-signin-tool/src/SignInAutomation.ts
saskliutas: Not sure what are the benefits of guessing narrower type in this case. The only goal of this API is to match the minimal surface of playwright API that tools inside this package use. This is needed so that consumer could pass `Page` from different `playwright` version without type errors. Because `Page` is created and passed down by the consumer in my opinion it makes sense to mark these returns as `Promise<unknown>` because we don't care about what is actually returned.

If we try to guess and narrow these return types it might start producing errors if `playwright` decides to change return type in newer version: [example](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgCoQM5gPICMBWAMsANYoDeAsAFDJ3IDmEYAygPYC2zAFqAwFzIAFAEpkAXgB8yAApROwDBAA8ANzbAAJpJoBfGjTgYAniATIYAVzNhgbEMkhYAkhwAOAGyFsCg9FjwiUggxKlp6OAB3OGAwZB98ADomVk4ePlEAbj0DaicwV08hMPpGZnYuMF4QAWQjU3NRCWkwKEsIPREgA)

change/@itwin-oidc-signin-tool-f94d62ff-891e-4f65-a154-dd105ce33a9b.json
saskliutas: Cleaned up and shortened

packages/oidc-signin-tool/src/SignInAutomation.ts
saskliutas: Removed

hl662 (approved):
Anmol has given the right comments so I'll defer to him for final approval

packages/oidc-signin-tool/src/SignInAutomation.ts
anmolshres98: playwright seems like a stable enough pkg to not switch up return types on a whim on any old release. I think this is fine though so resolving

packages/oidc-signin-tool/src/SignInAutomation.ts
anmolshres98: I'm not seeing a change on this part of the file? did you forget to push?

change/@itwin-oidc-signin-tool-f94d62ff-891e-4f65-a154-dd105ce33a9b.json
anmolshres98: i think you forgot to push this change? diff looks same?

saskliutas: My bad. Should be update now...

packages/oidc-signin-tool/src/SignInAutomation.ts
saskliutas: My bad. Should be updated now...

anmolshres98 (approved):

PR #352, opened 2026-07-16
feat(node-cli): add optional tokenEncryption hook and fix public-data-derivable cache key

anmolshres98:
The refresh-token cache in `@itwin/node-cli-authorization` encrypted its cache using a key derived entirely from public data (clientId + issuer URL). anyone who knew those values could reconstruct the key and decrypt a stolen cache file.

**Changes:**

- added an optional `tokenEncryption` hook on `NodeCliAuthorizationConfiguration` so consumers can supply their own encrypt/decrypt instead of the built-in cipher.
- the cipher key is now a random 32-byte value generated once per install and persisted to a key file (`0o600`, owner-only) alongside the cache.
- the key file is only trusted if it's a regular file (never a symlink), owned by the current user, with no group/other access, and the right length. anything else is discarded and regenerated.
- key file creation uses `O_CREAT | O_EXCL` so it can't be tricked into following a symlink or overwriting an existing file.
- old or corrupted cache entries now fail closed as a clean cache miss (re-auth) instead of throwing.

**Known limitation**

on Windows, POSIX file permission checks don't apply, so the key file protection is weaker there. consumers needing a stronger guarantee on Windows should use the new `tokenEncryption` hook.


change/@itwin-node-cli-authorization-422926f5-9a02-4a87-927c-170748bafa40.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added an optional tokenEncryption hook (e.g. Electron's safeStorage) for custom encryption; fixed refresh token cache encryption key being derivable from public data.",
+  "packageName": "@itwin/node-cli-authorization",
+  "email": "anmolshres98@users.noreply.github.com",
+  "dependentChangeType": "patch"
```
ben-polinsky, line 6: minor perhaps?

packages/node-cli/src/TokenStore.ts
```diff
@@ -48,34 +69,153 @@ export class TokenStore {
...
+  /** Ensures the store directory exists and (POSIX only) is owner-only. */
+  private ensureSecureStoreDirectory(): void {
+    if (!existsSync(this._dir))
+      mkdirSync(this._dir, { recursive: true, mode: 0o700 });
+
+    if (process.platform !== "win32") {
+      try {
+        chmodSync(this._dir, 0o700);
+      } catch {
+        // May not own the directory - silently move on.
```
MichaelSwigerAtBentley: Should we log a trace here?

packages/node-cli/src/TokenStore.ts
```diff
@@ -48,34 +69,153 @@ export class TokenStore {
...
+      return undefined;
+    }
+
+    const key = this.isTrustedKeyFileStat(stat) ? readFileSync(keyFilePath) : undefined;
+    if (key?.length === 32)
+      return key;
+
+    try {
+      unlinkSync(keyFilePath);
+    } catch { }
```
MichaelSwigerAtBentley: Same here, trace could be useful for debugging in the future

packages/node-cli/src/TokenStore.ts
```diff
@@ -48,34 +69,153 @@ export class TokenStore {
...
+
+    if (this._tokenEncryption) {
+      const encryptedBuffer = await this._tokenEncryption.encrypt(plaintext);
+      return { encryptedCache: encryptedBuffer.toString("base64") };
+    }
+
     const iv = randomBytes(16);
-    const key = createSecretKey(new Uint8Array(this.generateCipherKey()));
+    const key = createSecretKey(new Uint8Array(this.getCipherKey()));
     const cipher = createCipheriv("aes-256-cbc", key, new Uint8Array(iv));
```
MichaelSwigerAtBentley: I'm not sure about this, since I am unfamiliar, but my Ai recommended this:

[Medium] Built-in cipher is CBC, not AEAD.  aes-256-cbc  provides confidentiality but no integrity/authentication. For brand-new crypto code, prefer  aes-256-gcm : it authenticates the ciphertext (detects tampering/bit-rot deterministically instead of relying on JSON.parse failing), avoids CBC padding-oracle concerns entirely, and is the modern default. The stored shape already carries an  iv ; add an auth tag field. Since the threat model is "another local user reading a stolen cache," this isn't a blocker, but it's the one substantive thing I'd change before this becomes the reference implementation others copy.

Might be worth considering if we are already updating this function

anmolshres98: I am not super familiar either but after consulting with my ai, it thinks it's worth making this change. I also asked for ELI5 of this and I think it is a good explanation:
```text
The lockbox story

Imagine your secret note (the refresh token) is kept in a lockbox (encryption).

CBC (what we have now) is a lockbox that only does one thing: makes the note unreadable without the key. But here's the catch — it has no way of telling whether someone messed with the box while you were away. If someone shakes it, pokes holes in it, or swaps parts of the scrambled note inside, the box happily opens later and hands you... something. Maybe garbage, maybe a note that's been subtly changed. You only find out it's wrong when you try to read it and it looks like gibberish — and "it looks like gibberish" is a pretty flimsy way to detect tampering.

GCM (the new one) is the same lockbox, but with a tamper-evident seal — like the foil seal on a medicine bottle. When you lock the box, it stamps a special seal (the "auth tag") that's mathematically computed from everything inside. When you open it, the box first checks the seal. If even one bit was changed — by an attacker, a corrupted disk, a bad backup — the seal won't match and the box says loudly and immediately: "this was tampered with, refusing to open." No guessing, no "hmm this looks like gibberish."
```

packages/node-cli/src/TokenStore.ts
anmolshres98: changed

packages/node-cli/src/TokenStore.ts
anmolshres98: changed

anmolshres98:
ci failing cause of cves, fixing in separate pr here: https://github.com/iTwin/auth-clients/pull/353

packages/node-cli/src/TokenStore.ts
ben-polinsky: Ah had a review I didn't submit :(

Anyway, sounds worth it, but just ensure compatibility across platforms. 

Shame my sol or fable agent didn't catch this.

change/@itwin-node-cli-authorization-422926f5-9a02-4a87-927c-170748bafa40.json
anmolshres98: I am guessing you are asking about the versioning of node-cli itself and not other pkgs that depend on it. In which case, the `"type": "minor"` on ln 2 bumps this version by a minor after this change. Wasn't entirely sure what "dependentChangeType" does so had to do some [research here](https://microsoft.github.io/beachball/concepts/bump-algorithm.html#dependent-version-bumping) and it seems like the value of this doesn't really matter for this lib because no other pkg in this monorepo depend on it anyways.

packages/node-cli/src/TokenStore.ts
anmolshres98: I think it works fine across different platforms, from my understanding, because it uses `node:crypto` so wherever `node` works, this should work

change/@itwin-node-cli-authorization-422926f5-9a02-4a87-927c-170748bafa40.json
ben-polinsky: My bad, I misread. Was referring to type... you are good thanks

MichaelSwigerAtBentley (approved):

PR #354, opened 2026-07-30
Cache OIDC configuration for Electron client

alexdunae:
The [OIDC client configuration endpoint](https://ims.bentley.com/.well-known/openid-configuration) rarely changes and includes a `cache-control` header indicating it's cacheable for 24 hours.

Fetching the configuration values from this endpoint takes ~500-800ms, right in the initial sign-in code path.

- Uses `fetch` for OIDC config so we can get response headers
- Stores the config in Electron safeStorage
- Validates URLs coming from safeStorage to ensure they are valid Bentley endpoints
- Respects `Cache-Control` header values

While this won't have any impact on first-runs of applications (since their cache will be empty) it will improve the performance of auth on subsequent app opens throughout the day.

### Benchmark

Mode | Samples | Mean | p50 | p95
-- | -- | -- | -- | --
Uncached | 10 | 458.81 ms | 468.74 ms | 515.14 ms
Cached | 10 | 4.71 ms | 4.65 ms | 4.77 ms

### DIFF
--- a/packages/electron/scripts/verify-flow.cjs
+++ b/packages/electron/scripts/verify-flow.cjs
@@ -1,0 +1,56 @@
+// Run from the repository root:
+// pnpm --filter @itwin/electron-authorization build
+// IMJS_OIDC_CLIENT_ID=native-... IMJS_OIDC_ISSUER_URL=https://qa-ims.bentley.com pnpm --filter @itwin/electron-authorization exec electron scripts/verify-flow.cjs
+
+const os = require("node:os");
+const path = require("node:path");
+const { Logger, LogLevel } = require("@itwin/core-bentley");
+const { app } = require("electron");
+const { ElectronMainAuthorization } = require("../lib/cjs/ElectronMain");
+const { electronAuthLoggerCategory } = require("../lib/cjs/common/constants");
+
+Logger.initializeToConsole();
+Logger.setLevel(electronAuthLoggerCategory, LogLevel.Trace);
+
+const clientId = process.env.IMJS_OIDC_CLIENT_ID;
+if (!clientId) {
+  console.error(
+    "Set IMJS_OIDC_CLIENT_ID to your Native application client ID.",
+  );
+  process.exitCode = 1;
+  return;
+}
+
+const issuerUrl = process.env.IMJS_OIDC_ISSUER_URL ?? "https://ims.bentley.com";
+const redirectUri =
+  process.env.IMJS_OIDC_REDIRECT_URI ?? "http://localhost:3000/signin-callback";
+const scopes = process.env.IMJS_OIDC_SCOPES ?? "itwin-platform";
+const tokenStorePath =
+  process.env.IMJS_AUTH_CACHE_DIR ??
+  path.join(os.tmpdir(), "itwin-auth-clients", "electron-flow");
+
+async function main() {
+  await app.whenReady();
+  console.log(`Cache: ${tokenStorePath}`);
+
+  const client = new ElectronMainAuthorization({
+    clientId,
+    issuerUrl,
+    redirectUris: [redirectUri],
+    scopes,
+    tokenStorePath,
+  });
+
+  await client.signIn();
+  const accessToken = await client.getAccessToken();
+
+  if (!accessToken.startsWith("Bearer "))
+    throw new Error("Authorization completed without a bearer token");
+}
+
+main()
+  .catch((error) => {
+    console.error(error);
+    process.exitCode = 1;
+  })
+  .finally(() => app.quit());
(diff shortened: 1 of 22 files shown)


ben-polinsky:
Thanks!

> This caching would also make sense in the node-cli client. I figured I'd post this PR here for discussion and then decide if we want to expand the scope to include node-cli. Browser client already benefits from built-in browser caching.

I'd argue yes.

Now we touch on the internal shared package issue :)

alexdunae:
@ben-polinsky I did a little draft adding the equivalent OIDC cache to node-cli - it's a fair amount of duplicated code, as expected.

Is there any appetite for creating an `@itwin/auth-common` package where we could put shared code? It would mean publishing a new package.

There are some other candidates that could go into our hypothetical shared package in the future:

- scope matching, e.g. `scopesMatch`
- token expiration calculations

The other option is adding an esbuild or similar step. 

Do you think either of these is worth it (duplicate code / new package / new build step)?  I'm unsure, since my initial goal was startup performance for Electron apps.

ben-polinsky:
> Is there any appetite for creating an `@itwin/auth-common` package where we could put shared code? It would mean publishing a new package.

We haven't wanted to in the past, but it's starting to make some sense. 

> The other option is adding an esbuild or similar step.
> 
> Do you think either of these is worth it (duplicate code / new package / new build step)? I'm unsure, since my initial goal was startup performancet's st for Electron apps.

I can raise an issue about code duplication and maintainers can discuss options there. Hold our noses for the duplication in this PR for the time being?

toddsouthenbentley:
I think it would be possible to create a new package in this monorepo that isn't published and just privately consumed by those that need it.

alexdunae:
> I think it would be possible to create a new package in this monorepo that isn't published and just privately consumed by those that need it.

I'm pretty sure that requires an esbuild / bundling step.  So totally viable, but adds complexity.

packages/electron/scripts/verify-flow.cjs
alexdunae, line 1: I found these scripts helpful to actually exercise the changes, but we can remove if they don't seem useful

alexdunae:
I've updated this with the duplication in node-cli -- it's not terrible, but obviously not ideal either.  At least it can be a discussion point for a shared package in the future.

packages/electron/scripts/verify-flow.cjs
ben-polinsky: Might be worth it to throw them in a test for easier execution?

alexdunae: I considered that - node-cli currently doesn't have integration tests so it would be a bit of work to get them setup.  They also are more like manual verification steps and don't totally feel test-like to me. 

I added a script runner .... `pnpm verify:flow` to make things more ergonomic and official™️ - WDYT?

But... happy to purse the integration tests if you feel they would be better.

packages/node-cli/src/Client.ts
```diff
@@ -12,28 +12,43 @@ import { readFileSync } from "fs";
...
+  TokenResponse,
 } from "@openid/appauth";
+import { NODE_CLI_AUTH_LOGGER_CATEGORY } from "./Constants";
 
 /**
  * Logger category used in this package
  * @public
  */
-export const NODE_CLI_AUTH_LOGGER_CATEGORY = "node-cli-auth";
+export { NODE_CLI_AUTH_LOGGER_CATEGORY } from "./Constants";
```
ben-polinsky: nit: Do you need to specify the path if you've already imported the constant?

ben-polinsky (approved):

packages/electron/scripts/verify-flow.cjs
ben-polinsky: Agreed, that's sufficient.

packages/node-cli/src/Client.ts
alexdunae: good call, simplified to `export { NODE_CLI_AUTH_LOG }`

ben-polinsky:
Alright, so docs are annoying:

> Topic constants does not have a @docs-group-description in electron-authorization.ts. Please add a description.
 Topic OidcDiscoveryCache does not have a @docs-group-description in electron-authorization.ts. Please add a description.
 Topic Constants does not have a @docs-group-description in node-cli-authorization.ts. Please add a description.
 Topic OidcDiscoveryCache does not have a @docs-group-description in node-cli-authorization.ts. Please add a description.

Should be able to ask copilot to handle this. The basics of it are that any new file needs to have:
```ts

/** @packageDocumentation
 * @module Renderer
 */
 
 ```
Then see the `docs-index.md` file(s). This is annoyingly required for iTwin.js docs. The module has to match up with a docs-package-description. So you can apply one of the existing groupings (Renderer/Main/etc) or create one of your own.

As for the e2e tests, would have to look into why they are failing now.

alexdunae:
@ben-polinsky turns out integration tests were failing because we're so fast now ;)

I think we're pretty solid now.  Is this good to merge from your POV? Should I get some more review?

ben-polinsky:
> @ben-polinsky turns out integration tests were failing because we're so fast now ;)
> 
> I think we're pretty solid now. Is this good to merge from your POV? Should I get some more review?

I think we're good :)

